Issue #25, opened 2020-02-27
allow '**' in ignore files 


broeder-j:
Feature request, nice to have.

It would be very nice if one could allow for bash regs in the ignorefile list.
example 
```
ignore_files=['cdn**'] # to ignore files with names cdn01, cdn02, cdn23, ...
```
(if a code produces an arbitray number of files with certain naming pattern)

Otherwise thanks for your work! Currently everything is working nicely on linux. On mac I had an issue, which I did not yet managed to track down yet.

greschd:
Should be easy enough to do, using the [fnmatch.fnmatchcase](https://docs.python.org/3.5/library/fnmatch.html#fnmatch.fnmatchcase) to match filenames here: https://github.com/aiidateam/aiida-testing/blob/develop/aiida_testing/mock_code/_cli.py#L54

Do you want to do this?

broeder-j:
I can do it.

greschd:
Thanks! Feel free to open a separate issue for the mac problem - maybe it's to do with the use of environment variables to pass arguments to the executable?

greschd:
Closed in #26.

PR #26, opened 2020-02-27
allowed for expressions in ingnore file lists, implements issue #25, …

broeder-j:
…also adjusted test to include this

### DIFF
--- a/aiida_testing/mock_code/_cli.py
+++ b/aiida_testing/mock_code/_cli.py
@@ -18,18 +18,6 @@ from ._env_keys import EnvKeys
 SUBMIT_FILE = '_aiidasubmit.sh'
 
 
-def check_expression(filename, expression_list):
-    """returns True if 'filename' matches any expression in 'expression_list'"""
-
-    match = False
-    for expr in expression_list:
-        if fnmatch.fnmatch(filename, expr):
-            match = True
-            break
-
-    return match
-
-
 def run() -> None:
     """
     Run the mock AiiDA code. If the corresponding result exists, it is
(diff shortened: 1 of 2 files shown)


aiida_testing/mock_code/_cli.py
```diff
@@ -51,8 +64,8 @@ def run() -> None:
                 continue
             os.makedirs(os.path.join(res_dir, dirname), exist_ok=True)
             for filename in filenames:
-                if filename in ignore_files:
-                    continue
+                if check_expression(filename, ignore_files):
```
greschd: This could be done simpler:

```
if any(fnmatch.fnmatchcase(filename, expr) for expr in ignore_files):
```

greschd (changes requested):
Thanks! I have just one suggestion:

aiida_testing/mock_code/_cli.py
broeder-j: Sure, you are right. The increased computational complexity does not matter in that case.
I would go with ```fnmatch.fnmatch``` instead of ```fnmatch.fnmatchcase``` since this works on Windows, for unix and mac it does not matter, see 'https://docs.python.org/3.5/library/os.path.html#os.path.normcase'

greschd: You're right, `fnmatch` is the better choice.

Just out of curiosity, how is the computational complexity different? AFAIK, ``any`` on an iterator aborts when the first ``True`` is encountered.

greschd:
If you're wondering why the checks are not running, GitHub is apparently having some issues.

greschd (approved):

PR #27, opened 2020-03-11
update readme

ltalirz:


greschd (approved):

greschd:
Looks like I was a bit too quick with merging - right now the `LICENSE.txt` is Apache2 (same in `setup.json`). But, it probably makes sense to be consistent with other aiidateam projects and change to MIT. @ltalirz I'll leave the decision up to you.

ltalirz:
Yes... there may be good reasons to prefer Apache over MIT for personal projects, but if this eventually gets integrated into aiida-core it's just going to make things unnecessarily difficult.
I'll change over to MIT.

PR #28, opened 2020-03-11
start documenting mock_code

### LINKED ISSUE #18: Add documentation for mock_code
Add an introductory user guide for the `mock_code` submodule in the user guide.

ltalirz:
fix #18 

So far, I haven't even yet had the time to try it myself ;-)
But that's how I understand it should be used...


greschd (approved):
Thanks, that looks very nice!

One thing I just noticed is that I haven't completely thought through how the ``.aiida-testing-config.yml`` should / shouldn't be committed into the repository - will open a separate issue on that.

PR #29, opened 2020-03-12
First Implemetation of export cache functionality

### LINKED ISSUE #20: Add 'export_cache' implementation

broeder-j:
Hi,
I put this here for review, @PhilippRue  and @greschd since my time to work on this is sparse.
The code might be not totally fine yet, I left both fixture versions (run_with_cache, with_export_cache) side by side. Maybe we should do a first merge, that we can continue to work on smaller things at a time.

After the slack discussion:
The fixture run_with_cache was split, that  one can use export_cache and load_cache alone. There is are simple test for these. 
In order for the run_with_cache fixture test to work, I monkey patched code.get_object_to_hash and calcjob.get_object_to_hash (since it used the computer uuid). there is a fixture for this. thanks greschd for this suggestion. 

PhilippRue has solved this issue different by introducing the aiida-core and mock code with a session scope instead of function and imports of computers and codes. In aiida-kkr he has running examples of with run_with_cache. This may not be a solution for several developers.

- The builder hash now supports nested namespaces.

Further I have implemented also a second fixture to run with exports where you have to specify the export to use, feature suggestions to this by greschd:
- allow for multiple calculation classes  (currently all or one, maybe with_caching in aiida-core has to support it first)
- making sure computer/ code hash to the same value for different machines (should work with the monkeypatch fixture, there is a test for this)

Known issue:
Sometimes the caching when running does not work (which causes the tests to fail), which I do not understand the hash of the imported and running calcjob is the same. @greschd maybe you spot the reason for this.
- Then the fixture run_with_cache should also work with yield 
- maybe the diff workchain for testing should not be in the test file?

I left both fixture versions in, since I would like the export name to be choosen in an automatic way per default. fix #20


broeder-j:
ok the calcjob hash on the CI is different...

greschd:
Thanks! I will give this a more thorough review later, but regarding the first point: We can definitely merge this even if it's not 100% ready yet, since the same is true for the package as a whole.

Let me know if you want me to take a look at the CI issue.

broeder-j:
Also I am not sure how to resolve these pre-commit mypy issues:
```
tests/export_cache/test_export_cache.py:23: error: Class cannot subclass 'WorkChain' (has type 'Any')
tests/export_cache/test_export_cache.py:23: error: Base type WorkChain becomes "Any" due to an unfollowed import
aiida_testing/export_cache/_fixtures.py:315: error: Need type annotation for 'union_pk' (hint: "union_pk: Set[<type>] = ...")
Found 3 errors in 2 files (checked 13 source files)
```

tests/export_cache/test_export_cache.py
```diff
@@ -0,0 +1,195 @@
...
+from aiida.orm.querybuilder import QueryBuilder
+
+DiffParameters = DataFactory('diff')
+DiffCalculation = CalculationFactory('diff')
+CALC_ENTRY_POINT = 'diff'
+
+#### diff workchain for basic tests
+
+
+class DiffWorkChain(WorkChain):
```
greschd: Add a ``# type: ignore`` comment here. The reason for this ``mypy`` issue is that ``aiida-core`` does not have type annotations, so it gets confused when you try to subclass from ``WorkChain``.

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,327 @@
...
+
+        # This is executed after the test
+        if not cache_exists:
+            # TODO create datadir if not existent
+
+            # in case of yield:
+            # is the db already cleaned?
+            # since we do not the stored process node we try to get it from the inputs,
+            # i.e to which node they are all connected, with the lowest common pk
+            union_pk = ty.Set()
```
greschd: This should probably be ``union_pk : ty.Set[int] = set()``. The ``typing`` classes are usually not for instantiating, but for the type annotations only.

Here the problem is that ``mypy`` cannot figure out which type the _content_ of the set has, because it is empty. So it needs help, which the ``ty.Set[int]`` provides.

greschd:
I've added comments on how to fix the ``mypy`` issues - but I'm also not 100% convinced if ``mypy`` is a net benefit here, given that I also have limited experience with it so far.

This depends on whether you (and the other contributors) want to get acquainted with ``mypy``, or if it's mostly a nuisance. 

Tagging @PhilippRue @ltalirz for comment.

ltalirz:
Up to you, guys. I've never used mypy, but now that python 3.5 is the lowest python version we support, and 3.6 will be it in September, I would say it's only a matter of time for type hints to enter aiida-core (they would certainly be welcome!).

Since there seem to be a couple of different flavors (+ changes again in 3.6), I'd appreciate it if someone who has experience would write a short AEP on how we should go about this.

greschd:
Well, then maybe `aiida-testing` is a decent testing ground for that, too.

With "different flavors", do you mean external libraries? In the standard, I'm only aware of the `typing` module - and newer Python versions just add to that (and _where_ hints are allowed).

ltalirz:
Right, I just meant the different places for type hints - there is PEP 483, 484, 526, 3107 etc. - not external libraries.
I guess it's quite straightforward if you've used it once.

greschd:
Oh, I see. The way I understand it so far is these PEPs mostly build upon each other.. so there's no competition between flavors, just different levels of what's possible.

broeder-j:
concerning mypy. it would be nice if one could ignore all types injected from external libary imports. Which is the only think I find annoying. Also the type annotations make things less readable in my opinion. Through I see the benefits.

greschd:
The external library imports can be ignored by adding an entry in https://github.com/aiidateam/aiida-testing/blob/develop/mypy.ini, for each import that should be ignored.

Ignoring _all_ external imports is explicitly discouraged by `mypy`, because it can hide actual errors.

aiida_testing/export_cache/_fixtures.py
```diff
@@ -150,9 +157,10 @@ def _load_cache(path_to_cache=None, node=None, load_all=False):
         # for this we rehash all process nodes
         # this way we use the full caching mechanism of aiida-core.
         # currently this should only cache CalcJobNodes
-        qb = QueryBuilder()
-        qb.append(ProcessNode, tag='node')  # query for all ProcesNodes
-        to_hash = qb.all()
+        qub = QueryBuilder()
+        qub.append(ProcessNode)  # query for all ProcesNodes
+        qub.append(Code)
```
PhilippRue: This line breaks the cachability for a test in aiida-kkr: https://github.com/JuDFTteam/aiida-kkr/blob/bf474a2e37d4a54a0a38f8dbe96769847637c5bf/aiida_kkr/tests/calculations/test_vorocalc.py#L80

If I comment out `qub.append(Code)` it runs as expected.

aiida_testing/export_cache/_fixtures.py
```diff
@@ -81,6 +81,13 @@ def _export_cache(node, savepath, overwrite=True):
         :param overwrite: bool, default=True, if existing export is overwritten
         """
         from aiida.tools.importexport import export
+        # we rehash before the export
+        qub = QueryBuilder()
+        qub.append(ProcessNode)  # query for all ProcesNodes
+        qub.append(Code)
+        to_hash = qub.all()
+        for node1 in to_hash:
+            node1[0].rehash()
```
PhilippRue: probably not needed (commenting this out in my test case made no difference)

PhilippRue (commented):
I tested the `run_with_cache` fixture that takes the a builder as input and it works with a minor change (see below).

Thanks to the monkeypatch for the code hashing I was able to remove the export file for the code and use just the `aiida_local_code` fixture.

broeder-j:
ok towards that the hash on the CI differs:
objects to hash:
```
 [{'withmpi': False, 'resources': {'num_machines': 1, 'num_mpiprocs_per_machine': 1}, 'append_text': '', 'parser_name': 'diff', 'prepend_text': '', 'output_filename': 'patch.diff', 'scheduler_stderr': '_scheduler-stderr.txt', 'scheduler_stdout': '_scheduler-stdout.txt', 'mpirun_extra_params': [], 'environment_variables': {}, 'import_sys_environment': True, 'custom_scheduler_commands': ''}, {'file2': 'f744e4551095439a50a78488b89bf85e6b9bd347d5f99bbfd2fde64954cf91c5', 'file1': 'f6d1c358fb7a03d0831faf710b39ea945911fddab259be85ed0d25a51e62c992', 'parameters': '57b7298bdbd24a13add102c387767d645f9c8e0a86a9d0b1ab6492bda59e0758', 'code': 'bce6cfbda7e548b4af0f27cf5cd39a1245b21f3f6c85a0b5a9278ccc7f6837cc'}]
```
```
[{'withmpi': False, 'resources': {'num_machines': 1, 'num_mpiprocs_per_machine': 1}, 'append_text': '', 'parser_name': 'diff', 'prepend_text': '', 'output_filename': 'patch.diff', 'scheduler_stderr': '_scheduler-stderr.txt', 'scheduler_stdout': '_scheduler-stdout.txt', 'mpirun_extra_params': [], 'environment_variables': {}, 'import_sys_environment': True, 'custom_scheduler_commands': ''}, {'file2': 'a403c9809fc94ab203da6536a184e194a7856d50e73369345e26bb9ab1f39e5a', 'file1': 'ee69f1c4e4caefadf18fec511d6c780dbfdbcc7ad8b8bfcb4a954b3851b0f654', 'parameters': 'a5045df6d3d8b067897494f3facf13b340316569c13ac747b1c611fbf7e19725', 'code': 'bce6cfbda7e548b4af0f27cf5cd39a1245b21f3f6c85a0b5a9278ccc7f6837cc'}]
```
The parameters and the singlefile have a different hash on the CI and my computer. @greschd why?
is the abs path included in the hash for single file nodes, or is the repo location included for all nodes?

greschd:
Hmm, I'm not sure about this.. 

I think the abspath is not responsible, because you can ``.clone()`` a ``SingleFileData`` also locally, and get the same hash.
The repository location _shouldn't_ be in the hash, but I have never done testing across different AiiDA installations / profiles.

In principle there could be a problem with uppercase / lowercase file names (depending on the file system) due to the Folder hashing, which is implemented here: https://github.com/aiidateam/aiida-core/blob/develop/aiida/common/hashing.py#L260

I guess a helper to figure out which parts of the hash would be quite convenient right now (see https://github.com/aiidateam/aiida-core/issues/2549), but for now: Can you also print ``_get_objects_to_hash`` of the files and parameters?

greschd:
Just tried creating the same `SinglefileData` on two different systems (with different repo locations), and I'm also getting the same hash.

PhilippRue:
On my system the tests work fine with these changes (compared to commit 97cc9c6  `type def and ignore for mypy, reverted previous change, excluded aiida version from calcjob hash (4 hours ago) <Jens Bröder>`) all tests pass. 
```
diff --git a/aiida_testing/export_cache/_fixtures.py b/aiida_testing/export_cache/_fixtures.py
index c8a4362..13ef977 100644
--- a/aiida_testing/export_cache/_fixtures.py
+++ b/aiida_testing/export_cache/_fixtures.py
@@ -159,7 +159,8 @@ def load_cache(hash_code_by_entrypoint):
         # currently this should only cache CalcJobNodes
         qub = QueryBuilder()
         qub.append(ProcessNode)  # query for all ProcesNodes
-        qub.append(Code)
+        # this breaks it:
+        #qub.append(Code)
         to_hash = qub.all()
         for node1 in to_hash:
             node1[0].rehash()
@@ -222,7 +223,8 @@ def hash_code_by_entrypoint(monkeypatch):
         """
         Return a list of objects which should be included in the hash of a Code node
         """
-        return [self.get_attribute(key='input_plugin'), self.get_computer_name()]
+        return [self.get_attribute(key='input_plugin')]
+        #return [self.get_attribute(key='input_plugin'), self.get_computer_name()]
 
     def mock_objects_to_hash_calcjob(self):
         """
diff --git a/tests/export_cache/caches/diff_workchain.tar.gz b/tests/export_cache/caches/diff_workchain.tar.gz
index ee313ac..795c096 100644
Binary files a/tests/export_cache/caches/diff_workchain.tar.gz and b/tests/export_cache/caches/diff_workchain.tar.gz differ
diff --git a/tests/export_cache/test_export_cache.py b/tests/export_cache/test_export_cache.py
index 98eabab..86c22c7 100644
--- a/tests/export_cache/test_export_cache.py
+++ b/tests/export_cache/test_export_cache.py
@@ -111,7 +111,8 @@ def test_mock_hash_codes(mock_code_factory, clear_database, hash_code_by_entrypo
         ignore_files=('_aiidasubmit.sh', 'file*')
     )
     objs = mock_code._get_objects_to_hash()
-    assert objs == [mock_code.get_attribute(key='input_plugin'), mock_code.get_computer_name()]
+    assert objs == [mock_code.get_attribute(key='input_plugin')]
+    #assert objs == [mock_code.get_attribute(key='input_plugin'), mock_code.get_computer_name()]
 
 
 @pytest.mark.timeout(60, method='thread')
@@ -153,7 +154,7 @@ def test_run_with_cache(
     #Test if cache was used?
     diffjob = node.get_outgoing().get_node_by_label('CALL')
     cache_src = diffjob.get_cache_source()
-    calc_hash_s = '833070207922ee3dccbc477db7bbbf871d9df68bef2624604144f76b0edc6f0a'
+    calc_hash_s = '0c048c64551bd0b099ed82a90ff2c36833a3f1d0620712e7059367a67dac5c91'
     calc_hash = diffjob.get_hash()
     assert calc_hash == calc_hash_s
     assert cache_src is not None
@@ -200,7 +201,7 @@ def test_with_export_cache(
     #Test if cache was used?
     diffjob = node.get_outgoing().get_node_by_label('CALL')
     cache_src = diffjob.get_cache_source()
-    calc_hash_s = '833070207922ee3dccbc477db7bbbf871d9df68bef2624604144f76b0edc6f0a'
+    calc_hash_s = '0c048c64551bd0b099ed82a90ff2c36833a3f1d0620712e7059367a67dac5c91'
     calc_hash = diffjob.get_hash()
     assert calc_hash == calc_hash_s
     assert cache_src is not None
```

PhilippRue:
Note that changing the monkeypatch for the code hash is needed if one first imports an export file that contains the code because this gets the `(Imported )` addition to the name

broeder-j:
I exclude the version now from most aiida data types. had to monkeypatch, because the version is doubled in there. now it works. ;-). Or do we wont the tests to fail if the aiida version changes? If I have done it right, for pacakge dependend datatypes which do not inherit from Dict, RemoteData, SinglefileData the version check is still in. Maybe this can be done smarter, also basetypes and arrays are not patched yet. Or one could patch node in general to always exlcude the version.

greschd:
Hmm, maybe ignoring the (aiida or plugin) version can be a separate fixture? That way, a plugin developer can decide - if you always want to have the fixture, all you need to do is create a "dummy" fixture which uses it, and has `autouse=True` set.

PhilippRue:
> Hmm, maybe ignoring the (aiida or plugin) version can be a separate fixture? That way, a plugin developer can decide - if you always want to have the fixture, all you need to do is create a "dummy" fixture which uses it, and has `autouse=True` set.

I agree, now the `hash_code_by_entrypoint` fixture does more than just change hashing for the code (e.g. change it for `SingleFileData` as well).

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+@pytest.fixture(scope='function')
+def with_export_cache(export_cache, load_cache):
+    """
+    Fixture to use in a with() environment within a test to enable caching within the with.
+    Requires to provide an absolutpath to the export file to load or export to.
+    """
+    @contextmanager
+    def _with_export_cache(data_dir_abspath, calculation_class=None):
```
PhilippRue: if we have a default value for the data dir in `run_with_cache` (line 299) should we have the same here?

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,375 @@
...
+            # since we do not the stored process node we try to get it from the inputs,
+            # i.e to which node they are all connected, with the lowest common pk
+            union_pk: ty.Set[int] = set()
+            for node in input_nodes:
+                pks = {ent.node.pk for ent in node.get_outgoing().all()}
+                union_pk = union_pk.union(pks)
+            if len(union_pk) != 0:
+                process_node_pk = min(union_pk)
+                #export data to reuse it later
+                export_cache(node=load_node(process_node_pk), savepath=full_import_path)
```
PhilippRue: This logic seems problematic. Suppose we prepare the inputs to a workflow with a `calcfunction` which then has a lower pk. Then the export would be done for the calcfunction and not the workflow.

Why don't we just use `export_cache(node=resnode, ...)`? Then we are sure to export the workflow.

PhilippRue:
Also we might want to provide the option for the `with_export_cache` to overwrite the output file. Now the export file is only written if no previous file is found. 

This could be helpful for debugging/developing workflows (see https://groups.google.com/forum/?hl=en#!topic/aiidausers/y-ekFBlLgsg from mailing list).

aiida_testing/export_cache/_fixtures.py
broeder-j: This was implement because I tried no to use run_get_node within the fixture. if we say it is ok that the workchain is run within the fixture (fixture should only prepare and clean up), this part becomes obsolete. This fixture is therefore like a monkeypatch for run_get_node

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,375 @@
...
+    return bui_hash, input_nodes
+
+
+####
+
+#### fixtures
+
+
+@pytest.fixture(scope='function')
+def export_cache(hash_code_by_entrypoint):
```
greschd, line 79: I'm not sure if this should be a fixture (same for `load_cache`), or just an internal helper function. Do we want users to use the bare `export_cache`?

aiida_testing/export_cache/_fixtures.py
greschd: I'd actually prefer not having a default in either of the two places.

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,375 @@
...
+        data_dir: optional
+            Absolute path of the directory where the exported workchain graphs are
+            stored.
+
+        ignore_nodes : list string, ignore input nodes with these labels/link labels to ignore in hash.
+        # needed?
+        """
+
+        cache_exists = False
+        if data_dir == 'data_dir':
```
greschd: What is this special case for? As far as I can tell, this doesn't actually change the behavior: Below, we use `pathlib.Path(full_import_path)`, which - if for example `full_import_path` is `data_dir/something.tar.gz` - is relative to `cwd` anyway, no?

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,375 @@
...
+            # we assume ProcessBuilder, since type(ProcessBuilder) is abc
+        #else:
+        #    raise TypeError(
+        #        'builder has to be of type ProcessBuilder if no process_class is specified'
+        #    )
+        name = label + str(process_class).split('.')[-1].strip("'>") + '-nodes-' + bui_hash
+        print(name)
+
+        # check existence
+        full_import_path = str(data_dir) + '/' + name + '.tar.gz'
```
greschd: The hard-coded slash could be problematic: To be platform-agnostic, this should be 

```python
cache_path = pathlib.Path(data_dir) / (name + '.tar.gz')
```

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,375 @@
...
+            Absolute path of the directory where the exported workchain graphs are
+            stored.
+
+        ignore_nodes : list string, ignore input nodes with these labels/link labels to ignore in hash.
+        # needed?
+        """
+
+        cache_exists = False
+        if data_dir == 'data_dir':
+            cwd = pathlib.Path(os.getcwd())  # Might be not the best idea.
```
greschd: Yeah, using `os.getwcd()` is definitely not great -- the issue is that one might run `pytest` from different directories (e.g., form the top-level module directory or inside the `tests` directory). 

In my opinion, both should always work. There are two approaches to this that I've seen:
- In the (top-level) `conftest.py`, make use of `__file__` to define a fixture that gives you a constant "anchor directory". See e.g. https://github.com/FrescolinoGroup/pyhdf5io/blob/master/tests/conftest.py#L17
- Use the location of the test that you are currently running, and define paths relative to that. This is what `pytest-datadir` uses.

Both are valid, depending on the use case (do you want to share data between tests, or have it only for a specific test / test directory. I think we should leave this choice up to the user.

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,375 @@
...
+        cache_path = pathlib.Path(full_import_path)
+        if cache_path.exists():
+            cache_exists = True
+
+        if cache_exists:
+            # import data from previous run to use caching
+            load_cache(path_to_cache=full_import_path)
+
+        # now run process/workchain whatever
+        with enable_caching():  # should enable caching globally in this python interpreter
```
greschd, line 359: Shouldn't it be possible to re-use `with_export_cache` here? That exports the calculations while here we explicitly export the parts connected to the node, but is this distinction really needed?

We could maybe change the `with_enable_caching` logic to only export calculations that were created _within_ the context manager (and their direct inputs / outputs). So, for example
```python
run(SomeCalculation, **<inputs>)
with with_export_cache(<...>):
    run(OtherCalculation, **<inputs>)
```
would export the `OtherCalculation`, but not `SomeCalculation`. That would also get around the need to always clean the DB before running the test.

aiida_testing/export_cache/_fixtures.py
PhilippRue: I don't have strong feelings towards one or the other. It should just be the same for both ways

aiida_testing/export_cache/_fixtures.py
PhilippRue: OK but this did not work for [this test](https://github.com/JuDFTteam/aiida-kkr/blob/develop/aiida_kkr/tests/workflows/test_kkrimp_sub_wc.py) where I prepare an input to the workflow with a `calcfunction` which therefore has a lower pk. Then only the `calcfunction`  is exported but not the workflow nodes itself.

But maybe this can be fixed with an option for the export.

aiida_testing/export_cache/_fixtures.py
broeder-j: Currently run_with_cache (which is kind of a patch for run_get_node) exists only because I do not know how to explicitly do this and I want an automatic cache naming which changes if the test changes, otherwise debugging becomes harder. How do you know what is run in the `with` environment? One idea is query for creation time. Then on the other hand if you use as we currently do the default aiida export you get the full provenance, which means if your `SomeCalculation` and `OtherCalculation` are connected (which within a test they usually are) they will get exported both anyway.

aiida_testing/export_cache/_fixtures.py
broeder-j: I will change this. I was not familiar with pathlib and why to use it at all and not os. Same goes for the tests

aiida_testing/export_cache/_fixtures.py
broeder-j: I made this an extra fixture, because if we change the import/export functionality to be different from what aiida-core does, as a user I might want to use it in a test (or see what it does for debugging). In aiida-fleur tests I use a fixture to export and import things to create test environments, because we also have some input nodes for workchains where uuids are provided. So there might be a need to have something `bare` which is adapted for tests (like with specific hashing, or striped/slim exports etc.). For now this defines the interface. Overall I'm also not sure if this should be just a helper as for my current purposes do not need to take it further.

aiida_testing/export_cache/_fixtures.py
broeder-j: You are right

aiida_testing/export_cache/_fixtures.py
broeder-j: we use the package file path for our plugins. I also did not like cwd. Since relative paths are to cwd, the default is ensured to work, which was the only point in doing so.

aiida_testing/export_cache/_fixtures.py
greschd: Yeah, I think it makes sense to just export what is created during the `with` statement (and its provenance). That can be either by creation time, or maybe by checking what exists already in the DB when entering the `with` statement - whatever works more reliably. 
I'm not sure if we might get into clock synchronization issues (between the DB and the python interpreter, maybe?) when using `ctime`.

broeder-j:
> Also we might want to provide the option for the `with_export_cache` to overwrite the output file. Now the export file is only written if no previous file is found.
> 
> This could be helpful for debugging/developing workflows (see https://groups.google.com/forum/?hl=en#!topic/aiidausers/y-ekFBlLgsg from mailing list).

Done

aiida_testing/export_cache/_fixtures.py
broeder-j: Comment: 'Spliting' the cache
works currently for both fixtures, since the export happens at the end of the with.
```
with with_export_cache(<...>):
    run(OtherCalculation, **<inputs>)
with with_export_cache(<...>):
    run(ThisWeCacheSeperateCalculation, **<inputs>)
```
So, if you know to extract what ran in the contextmanager let me know.

aiida_testing/export_cache/_fixtures.py
```diff
@@ -345,6 +345,8 @@ def _run_with_cache( # type: ignore
 
         # check existence
         full_import_path = pathlib.Path(data_dir) / (name + '.tar.gz')
+        # make sure the path is absolute (this is needed by export_cache)
+        full_import_path  = full_import_path.absolute()
```
PhilippRue: if this is not an absolute path `export_cache` fails because there this is checked

PhilippRue (dismissed):
I think all my complains are in now. I mainly tested the `run_with_cache` fixture which now works fine for me

PhilippRue (approved):

aiida_testing/export_cache/_fixtures.py
greschd: I think we should _check_ that it is an absolute path instead of just making it absolute. Otherwise the test will depend on which directory they are run from. 

As much as possible, e.g. running `pytest` from the root directory should do the same as e.g. `cd tests; pytest`.

PhilippRue: oh yes, that slipped my mind. so how about we make this a check only and get rid of the default value. Then the users would be required to set the paths to executables, output directory etc. themselves and we wouldn't be hiding this in the fixture.
I guess this would be cleaner.

aiida_testing/export_cache/_fixtures.py
greschd: I will have to do some testing to see if there is a good way to extract that...

greschd:
@broeder-j @PhilippRue I think we can do the following: We merge this PR now - I will go over the functionality and try to implement the "node selection by what ran in the context manager" discussed above. After that I will create a new PR (possibly with some interface changes...), and have the two of you review it / test it on your plugins.

Does that work for you?

PhilippRue:
> @broeder-j @PhilippRue I think we can do the following: We merge this PR now - I will go over the functionality and try to implement the "node selection by what ran in the context manager" discussed above. After that I will create a new PR (possibly with some interface changes...), and have the two of you review it / test it on your plugins.
> 
> Does that work for you?

sounds good 👍

greschd:
Since it seemed to be incompatible with the changes in #30, I merged it into a separate branch for now (to keep the tests running on `develop`).

PR #33, opened 2020-03-18
add cli options for config file management

### LINKED ISSUE #8: Add flag to force regenerating files
We should add a command-line flag to wipe the output directory before running the tests.

In the case of "uniquely labelled" content, maybe there should be an option to only regenerate it if the test was actually run.

Also useful would be a way to delete only outputs that were never touched during the test run.

### LINKED ISSUE #13: Add sanity check to see that 'aiida-mock-code' executable runs
Add a "sanity check" session-scoped fixture (with autouse) that checks that the ``aiida-mock-codes`` executable works. For example, it could fail when ``pkg_resources`` detects an incompatibility in the requirements.

This probably also needs some change in the ``aiida-mock-codes`` executable itself, giving it a ``--sanity-check`` option or similar.

### LINKED ISSUE #30: Enable running without a config file
Currently, the ``.aiida-testing-config.yml`` config file is committed to the repository. In general (and for plugins), this kind of defeats the purpose of the config file: It contains the parts that are _different_ on different systems, and can not be hardcoded in the tests.

A partial solution is to just recommend not committing ``.aiida-testing-config.yml``, however it would be good to make all fixtures run (if all caches are present) also _without_ a config file, for the purposes of CI. Having to create a special "CI config file" instead would be quite cumbersome.

ltalirz:
fixes #30 
fixes #8
fixes #13

Add cli option:

  --testing-config-action: Read config file if present ('read'), require config file ('require') or generate new config file ('generate').
  --regenerate-test-data: Regenerate test data

A few things left to do:
 - [x] add docs
 - [ ] use type hints more?


ltalirz:
Just to point out that some strange stuff was going on with github actions (see e.g. the ci results of "add test for regeneration of test data").

Also, on my computer, the 4 tests that involve running the CalcJob fail with error messages
```
$ cd tests; pytest

...

    def check_diff_output(result):
        """
        Checks the result from a diff calculation against a reference.
        """
        diff_res_lines = tuple([
            line.strip() for line in result['diff'].get_content().splitlines() if line.strip()
        ])
>       assert diff_res_lines == (
            "1,2c1", "< Lorem ipsum dolor..", "<", "---",
            "> Please report to the ministry of silly walks."
        )
E       AssertionError: assert () == ('1,2c1', '< ...silly walks.')
E         Right contains 5 more items, first extra item: '1,2c1'
E         Use -v to get the full diff
```
(i.e. check_diff is empty for those). Haven't figured out why yet.

ltalirz:
This seems to have been fixed by requiring the data directory to be an absolute path.

aiida_testing/mock_code/_fixtures.py
```diff
@@ -123,6 +123,10 @@ def _get_mock_code(
         # can not be reused.
         code_label = f'mock-{label}-{uuid.uuid4()}'
 
+        data_dir_abspath = pathlib.Path(data_dir_abspath).absolute()
```
greschd: Hmm, I'm not entirely sure we should be converting to an absolute path _here_. The reason I wanted `data_dir_abspath` to be _given_ as an absolute path is because resolving the path to absolute here will change depending on where the tests are executed from. For example, ``pytest`` from the project root will not do the same as ``cd tests; pytest``.

Maybe a better choice would be to _check_ that the path is absolute?

ltalirz: Absolutely. Fixed.

aiida_testing/_config.py
```diff
@@ -1,30 +1,87 @@
...
+        return self._dict.__getitem__(item)
+
+    def __setitem__(self, key, value):
+        return self._dict.__setitem__(key, value)
+
+    def __delitem__(self, key):
+        return self._dict.__delitem__(key)
+
+    def __iter__(self):
+        return self._dict.__init__()
```
greschd: I think this is a typo - it should be `__iter__`, not `__init__`.

aiida_testing/mock_code/_fixtures.py
```diff
@@ -8,33 +8,91 @@
...
+    "testing_config",
+    "mock_code_factory",
+)
+
+
+def pytest_addoption(parser):
+    """Add pytest command line options."""
+    parser.addoption(
+        "--testing-config-action",
+        type=click.Choice((c.value for c in ConfigActions)),
```
greschd, line 32: Neat, I didn't know you can use `click.Choice` in combination with `pytest` options 😃

aiida_testing/mock_code/_fixtures.py
```diff
@@ -8,33 +8,91 @@
...
+    """
     def _get_mock_code(
         label: str,
         entry_point: str,
         data_dir_abspath: ty.Union[str, pathlib.Path],
-        ignore_files: ty.Iterable[str] = ('_aiidasubmit.sh', )
-    ):
+        ignore_files: ty.Iterable[str] = ('_aiidasubmit.sh'),
+        executable_name: str = '',
+        config: dict = testing_config,
```
greschd: Are `config`, `config_action` and `regenerate_test_data` here passed just for being able to test the fixture internally, or do we expect users to set them?

I'm not really sure if there's an established convention for that, but maybe we could indicate this by a leading underscore in the kwarg names (`_config`, `_config_action`, `_regenerate_test_data`)?

aiida_testing/mock_code/_fixtures.py
```diff
@@ -48,26 +106,58 @@ def _get_mock_code(
...
-            when the code is executed.
+            after the code has been executed.
+        executable_name :
+            Name of code executable to search for in PATH, if configuration file does not specify location already.
+        config :
+            Dict with contents of configuration file
+        config_action :
+            If 'require', raise ValueError if config dictionary does not specify path of executable.
+            If 'generate', add new key (label) to config dictionary.
+        generate_config :
```
greschd: The `generate_config` parameter doesn't exist, and `regenerate_test_data` is missing in the docstring.

greschd (changes requested):
Very nice, thanks! I've added a few minor comments.

For regenerating the data, I think the "export cache" feature will need a similar flag, but they should probably be independent: Maybe we should rename  `--regenerate-test-data` to something like `--regenerate-mock-code-data`, to distinguish them?

aiida_testing/_config.py
ltalirz: thanks, well spotted!

aiida_testing/mock_code/_fixtures.py
ltalirz: done!

aiida_testing/mock_code/_fixtures.py
ltalirz: thanks, fixed!

greschd (dismissed):
Thanks! Do you want to create a new issue for the "add docs" part?

ltalirz:
happy to add them in here, now that we've agreed on the names

greschd:
> fixes #30
> fixes #8
> fixes #13
> 

Drei auf einen Streich! Nice 😄

PR #34, opened 2020-03-31
faster entry point loading for console script

ltalirz:
use fastentrypoints module [1] to load console scripts.
not only is it faster, the alternative route via pkg_resources also
throws exceptions whenever it detects dependency clashes in the python
environment, causing aiida-mock-code to fail.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
   rev: v0.740
   hooks:
   - id: mypy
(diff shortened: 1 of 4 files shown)


greschd:
Thanks! I think it needs to be added to the `MANIFEST.in` file as well, so that it gets included when building the package.

ltalirz:
right, done!

greschd (approved):

greschd:
Fantastic, thanks!

PR #38, opened 2020-10-26
modernized setup

### LINKED ISSUE #37: Fix / modernize install
As reported by @DanielMarchand in #36, the install can fail if `fastentrypoints` cannot be found.

This can be fixed by un-vendoring `fastentrypoints`, and introducing the `[build-system]` info in a `pyproject.toml`. While we're at it, we can also consider switching from `setup.json` with a "complex" `setup.py`, to `setup.cfg` with a minimal `setup.py`.

Example `setup.py`: https://github.com/Z2PackDev/TBmodels/blob/dev/setup.py  
Here we actually fail gracefully if `fastentrypoints` isn't present. I think this is for the purpose of running `python setup.py {sdist,bdist_wheel}`. Because `pip` isn't involved in that, the build isolation + environment setup doesn't work as expected.

Example `pyproject.toml`: https://github.com/Z2PackDev/TBmodels/blob/dev/pyproject.toml

Both these files can be used pretty much unaltered AFAICT, only `setup.cfg` would need adapting.

See #36 for additional discussion.

DanielMarchand:
Fixes #37.


setup.cfg
```diff
@@ -0,0 +1,57 @@
+[metadata]
+name = aiida-testing
+author = Dominik Gresch
```
greschd: While we're at it - I'm not the only author of this (anymore). Should at least include

```
Dominik Gresch, Leopold Talirz, Jens Bröder, and the AiiDA team
```

The `author_email` can be left blank I suppose.

greschd: (you can put yourself also if you want, of course)

setup.cfg
```diff
@@ -0,0 +1,57 @@
...
+author_email = greschd@gmx.ch
+version = 0.1.0.dev1
+url = https://aiida-testing.readthedocs.io/
+description = Helpers for testing AiiDA plugins using pytest.",
```
greschd: There's a straggling `",` here.

setup.cfg
```diff
@@ -0,0 +1,57 @@
+[metadata]
+name = aiida-testing
+author = Dominik Gresch
+author_email = greschd@gmx.ch
+version = 0.1.0.dev1
```
greschd: We should use `attr: aiida_testing.__version__` instead of hard-coding the version. Then we can also get rid of the `validate_version_consistency` pre-commit check.

greschd:
Please also remove the vendored `utils/fastentrypoints.py`.

greschd (approved):
Fabulous, thanks @DanielMarchand!

PR #41, opened 2020-10-26
Add non-editable install test to CI; test py39 instead of py38

### LINKED ISSUE #39: Add CI tests that install without -e
The current CI tests all install the package with `pip install -e .`. This doesn't fully test the packaging, because it roughly corresponds to adding the directory to `sys.path`. If there is a mistake in the manifest file or listed packages to install, it will not be caught.

We should add a job that installs with `pip install .`, without `-e`. 

Prompted by a regression introduced in #38, which is fixed by #40.

greschd:
Fixes #39.

Adds two variants of the 'tests' job in CI, installed with and without the '-e' option. 

Also replaces py38 with py39 as 'upper' Python version to test.

We can merge with admin privileges and then update the "required" checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,9 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python-version: [3.6, 3.8]
+        python-version: [3.6, 3.9]
         backend: ['django']
+        editable_install_option: ['-e', '']
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


ltalirz (dismissed):
thanks @greschd - looks good!

ideally one might be able to test the "pip install ." just on one of the two python versions but then one cannot use a matrix anymore

greschd:
Is there maybe some way to exclude specific matrix combinations?

*Update:* yes, there is..

greschd:
Debatable of course whether `-e` should be the one we test on both versions of Python, or the other way around. But I think it doesn't matter much - if there is something that breaks the install process, it's more likely to be related to the version of `pip` or `setuptools` rather than Python itself.

ltalirz (approved):
nice, thanks for teaching me about the jobs matrix :-)

PR #45, opened 2020-11-16
mock-code: add 'ignore_paths' argument

### LINKED ISSUE #44: mock-code: allow ignoring files based on path
The current code for the `ignore_files` parameter allows ignoring files only based on their filename, not based on their path in the the run directory

https://github.com/aiidateam/aiida-testing/blob/4aa5c3eaea1f4cb55f5a3d6be015e35867a3c77c/aiida_testing/mock_code/_cli.py#L55-L60

However, I have a use case where a code creates a bunch of folders and I simply want to ignore the content of some of these - e.g. I'd like to ignore `'VTK/*'`

An `ignore_paths` approach seems generally more powerful to me, so I propose to add this.

ltalirz:
fixes #44 

The `ignore_files` parameter of the mock-code fixture allows ignoring
files only based on their filename, but not based on their path in the
the run directory

This does not cover the use case when a code creates a subfolder
structure and one would like to ignore all files in certain subfolders.


ltalirz:
@greschd Not sure whether you noticed the review request (if yes, no problem - it's not urgent)

greschd:
Yep, just didn't get around to it yet :)

aiida_testing/mock_code/_fixtures.py
```diff
@@ -121,7 +129,21 @@ def _get_mock_code(
...
         _regenerate_test_data :
             If True, regenerate test data instead of reusing.
+
+        .. deprecated:: 0.1.0
+            Keyword `ingore_files` is deprecated and will be removed in `v1.0`. Use `ignore_paths` instead.
         """
+        if ignore_files != ('_aiidasubmit.sh', ):
+            warnings.warn(
+                'keyword `ignore_files` is deprecated and will be removed in `v1.0`. Use `ignore_paths` instead.',
+                AiidaDeprecationWarning
```
greschd: I'm not sure if we need to use `AiidaDeprecationWarning` here - `pytest` also shows plain `DeprecationWarning` by default.
But I guess it doesn't matter much either way, so we can also keep it.

tests/mock_code/test_ignore_paths.py
```diff
@@ -0,0 +1,70 @@
...
+        dest_dir=storage_directory,
+        ignore_files=(),
+        ignore_paths=('my/subfolder', )
+    )
+    assert (storage_directory / '_aiidasubmit.sh').is_file()
+    assert (storage_directory / 'file2.txt').is_file()
+    assert not (storage_directory / 'my' / 'subfolder' / 'file3.txt').is_file()
+
+    # ignore only specific file (from subfolder)
+    copy_files(
```
greschd, line 60: I think we should add a copy of this specific test where `ignore_paths` is used instead of `ignore_files`.

greschd (dismissed):
Looks great, thanks @ltalirz. Just some minor comments.

aiida_testing/mock_code/_fixtures.py
ltalirz: Thanks, I can fix this in one of the next PRs.

ltalirz: Done

tests/mock_code/test_ignore_paths.py
ltalirz: Thanks for the comment - this made me think that `rglob` is probably not the best solution here for `ignore_paths` (and I went back to `glob`).

I've added another test for `ignore_files` now; please have a look

greschd: That means once we remove `ignore_files` one would just have to use `**/filename` to ignore everything of a given name, right? I think that's fine (more explicit).

greschd (approved):
Let 'er rip! 🚀

PR #46, opened 2020-11-16
drop prospector dependency

ltalirz:
aiida-core moved away from prospector due to frequent dependency issues.
This PR replaces prospector by pylint-only, updates the pylint version
and moves the pylint configuration to the pyproject.toml file.


greschd (approved):
Looks good, thanks @ltalirz!

PR #47, opened 2020-11-19
Support MPI executables (on a single process)

ltalirz:
This builds on #45 and makes it possible to run executables through `mpirun`, but on a single-process only(!).

Previously, when the mock executable would detect that an input is not present in the cache, it would
 * replace its name in the submit script with the fallback executable
 * run the submit script (again) using bash

This approach obviously does not work in conjunction with MPI, since it ends up calling `mpirun` twice: once for running the mock executable, and once again _inside_ the mock executable when it calls the submit script again.

This PR changes the mock executable to simply call the fallback executable directly, using the command line arguments that were provided.
Environment variables are forwarded to the subprocess automatically using `subprocess.call`.

This works for running with `-np 1`, but it still does not solve running with more than one process (one should probably add a check).


ltalirz:
Reviving this pull request, as it is used in `aiida-lsmo` required by @mpougin

@greschd do you see any obvious issues with this approach?

aiida_testing/mock_code/_cli.py
```diff
@@ -46,8 +46,7 @@ def run() -> None:
             sys.exit("No existing output, and no executable specified.")
 
         # replace executable path in submit file and run calculation
```
greschd:
```suggestion
```

greschd (dismissed):

greschd:
LGTM, apart from the suggestion above (remove outdated comment).

ltalirz:
Thanks for the quick review @greschd, much appreciated!

Issue #51, opened 2022-02-18
mock-code: Allow code paths to be relative to config file location


janssenhenning:
In `aiida-fleur` we would like to place the executables to generate mock code test data relative to the location of the `.aiida-testing-config.yml`. Currently this is only possible with some hacks (See 
https://github.com/JuDFTteam/aiida-fleur/blob/develop/.github/workflows/ci.yml#L136
 for an example)
Here the paths are modified in the CI to be absolute
```
sed -i "s/\./${GITHUB_WORKSPACE//\//\\/}\/tests/g" .aiida-testing-config.yml
./run_all_cov.sh --local-exe-hdf5
```
We usually don't add compiled `fleur` executables to a global bin folder, since there are multiple executables with different configurations on the same machine very often.

greschd:
I don't see a downside to this, but won't have the time to implement it myself. Making the path absolute can probably happen after this code block

https://github.com/aiidateam/aiida-testing/blob/0d104fbad481c6dd2e5a0725e65dc2208fb8058b/aiida_testing/mock_code/_fixtures.py#L170-L174

For the config file path, I think it makes sense to add that to the config as a "special" key in `Config.from_file`, making sure to exclude it in `Config.to_file`.

Tagging @ltalirz for comment.

ltalirz:
Also no objections from my side.

I actually had the need to store also a `prepend_text` per code (see https://github.com/ltalirz/aiida-testing/tree/add-prepend-text), i.e. I needed to allow for several fields to be stored per code in the config file.
Anyhow, this is not needed for this feature.

PR #52, opened 2022-03-02
`mock-code`: Allow paths to be relative to the config file

janssenhenning:
See #51 

This PR allows paths in the `.aiida-testing-config.yml` file to be relative. These are interpreted to be relative to the config file location, since the location of the test execution can vary.

I followed the suggestion in the issue above and added the file path as a special key to the config, which is excluded when writing the config out. Another possibility would be to add the filepath as an attribute/property to the `Config` class. I don't have a preference for either way

### DIFF
--- a/aiida_testing/_config.py
+++ b/aiida_testing/_config.py
@@ -13,6 +13,7 @@ from voluptuous import Schema
 import yaml
 
 CONFIG_FILE_NAME = '.aiida-testing-config.yml'
+CONFIG_FILE_PATH_ENTRY = 'file_path'
 
 
 class ConfigActions(Enum):
@@ -52,6 +53,7 @@ class Config(collections.abc.MutableMapping):
             if config_file_path.exists():
                 with open(config_file_path, encoding='utf8') as config_file:
                     config = yaml.load(config_file, Loader=yaml.SafeLoader)
+                    config[CONFIG_FILE_PATH_ENTRY] = config_file_path
                     break
         else:
             config = {}
@@ -68,8 +70,10 @@ class Config(collections.abc.MutableMapping):
         cwd = pathlib.Path(os.getcwd())
         config_file_path = (cwd / CONFIG_FILE_NAME)
 
+        _dict = self._dict.copy()
+        _dict.pop(CONFIG_FILE_PATH_ENTRY)
         with open(config_file_path, 'w', encoding='utf8') as handle:
-            yaml.dump(self._dict, handle, Dumper=yaml.SafeDumper)
+            yaml.dump(_dict, handle, Dumper=yaml.SafeDumper)
 
     def __getitem__(self, item):
         return self._dict.__getitem__(item)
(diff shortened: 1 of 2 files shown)


greschd:
> Another possibility would be to add the filepath as an attribute/property to the `Config` class. 

That seems like the cleaner solution to me, since it doesn't have the "usual" properties of a config entry (e.g., cannot be missing, or changed via the usual means).

ltalirz (commented):
Hey guys, so am I understanding correctly that you are proposing to specify the filepath of the `.aiida-testing-config.yaml` inside the `.aiida-testing-config.yaml` file?
Meaning if something happens in the upstream directory structure, it will stop working?

Since pytest anyhow needs to find the `.aiida-testing-config.yaml` file (i.e. we know its location), wouldn't it be easier to 
 - check whether path to a code executable is absolute
 - if not, try treating it as a relative path inside the directory of the `.aiida-testing-config.yaml` file

?

greschd:
> Hey guys, so am I understanding correctly that you are proposing to specify the filepath of the `.aiida-testing-config.yaml` inside the `.aiida-testing-config.yaml` file?

No, AFAIU the goal is to provide the path at runtime in the `Config` (either as a direct attribute, or an entry in the dict), not hard-code it in the `.yaml`.

janssenhenning:
> Hey guys, so am I understanding correctly that you are proposing to specify the filepath of the `.aiida-testing-config.yaml` inside the `.aiida-testing-config.yaml` file? 

The filepath is never written out to the config file. In the current implementation it is added as a special key to the instance of the `Config` class in `from_file` and excluded again in `to_file`.

I will switch the file path to be a property of the `Config` class instead of a special key to make this distinction clearer.

ltalirz:
> I will switch the file path to be a property of the `Config` class instead of a special key to make this distinction clearer.

Thanks, I guess that's what confused me (besides being a bit tired ;-) ).

Sounds all fine; happy to merge once it's done

janssenhenning:
@ltalirz The file path is now implemented as a property
The tests and docs build fail on my fork but I think this is related to aiida-core 2.0 (reentry is now not available)/and a newer sphinx version complaining about the configuration

ltalirz:
Thanks @janssenhenning !
Let me have a look whether I can fix the CI

Are you using the package with AiiDA 1.6 or with AiiDA 2?

ltalirz:
I've fixed CI. Could you please update this branch? Somehow I don't have the rights to do so

image

janssenhenning:
> I've fixed CI. Could you please update this branch? Somehow I don't have the rights to do so

Done 👍 It's probably because I used a fork in our Institute organization, where the rights are probably different

ltalirz:
Thanks @janssenhenning !

Please note that one part of fixing CI was to make the limitation in the aiida-core version actually take effect, meaning it will force-install aiida<2 now.

However, I believe the package works with aiida2 as well - if you need that, let me know (or open a PR that removes the requirement and removes the `reentry` calls in the CI tests).

PR #56, opened 2022-07-18
New attempt at implementing `export_cache` fixture

### LINKED ISSUE #20: Add 'export_cache' implementation

janssenhenning:
Based on #29 

These are improvements on top of the initial `export_cache` fixtures that were written by me and @PhilippRue to be able to run `aiida-fleur` and `aiida-kkr` workflow tests with both aiida v1 and v2.

Notable changes from #29
- `load_cache`, `export_cache` and `hash_code_by_entrypoint` are cross-compatible between AiiDA versions 1.X and 2.X
- Cache path handling changes
    - If the path to `with_export_cache` etc. is relative, it is either interpreted relative to the default cache dir (see below) or if not specified to `Path(request.fspath.dirname) / 'caches'`, i.e. in the same directory as the test file
    -  if the path is absolute nothing changes
- Added flag `--export-cache-allow-migration` flag, which attempts to migrate archives if they are not compatible
- Added options to config file
    -  `default_data_dir`: if specified caches with relative paths are placed here
    - `calcjob_ignored_inputs/calcjob_ignored_attributes`: Additional things to exclude from calcjob hashes. Between v1 and v2 for example the metadata option `environment_variables_double_quotes` has to be ignored
    - `node_ignored_attributes`: dict, where data nodes (specified by entry point) can ignore additional attributes
- Fixed bug in monkeypatching Calcjob nodes. Here the version was not added to the ignored attributes by mistake 

The ignoring of things like `environment_variables_double_quotes` is not ideal, but I think being explicit with this is better for now.


ltalirz:
@janssenhenning are you still planning to complete this PR?

janssenhenning:
@ltalirz Yes I think this would be useful to finish up, since we already use it with a lot of success for veryfying that our plugins work with aiida 1.0 and 2.0

I cleaned the formatting up a bit and added some explanations to the now quite convoluted mokeypatching code
What I'm not quite sure on is how and whether I could write more fine-grained tests for these fixtures

As with #57 I have no idea what to do about the test failures

ltalirz:
Cheers! I'll try to find the time to have a look at this over the weekend

.aiida-testing-config.yml
```diff
@@ -1,2 +1,6 @@
 mock_code:
   diff: /usr/bin/diff
+export_cache:
+  calcjob_ignored_attributes:
+    - environment_variables_double_quotes
```
ltalirz, line 5: Let's add a comment here on why this is ignored (v1 vs v2)

.aiida-testing-config.yml
```diff
@@ -1,2 +1,6 @@
...
+    - submit_script_filename
```
ltalirz, line 6: same here

.github/workflows/ci.yml
```diff
@@ -51,7 +52,7 @@ jobs:
     - name: Run test suite
       env:
         # show timings of tests
-        PYTEST_ADDOPTS: "--durations=0"
+        PYTEST_ADDOPTS: "--durations=0 --export-cache-allow-migration"
```
ltalirz, line 55: I wonder whether this should anyhow be the default?

Perhaps the right approach would be to make this the default behavior and print a warning for any export that needs to be migrated.

aiida_testing/_config.py
```diff
@@ -26,7 +26,15 @@ class ConfigActions(Enum):
...
     """Configuration of aiida-testing package."""
 
-    schema = Schema({'mock_code': Schema({str: str})})
+    schema = Schema({
+        'mock_code': Schema({str: str}),
+        'export_cache': {
+            'default_cache_dir': str,
+            'calcjob_ignored_inputs': [str],
+            'calcjob_ignored_attributes': [str],
+            'node_ignored_attributes': Schema({str: [str]})
```
ltalirz, line 35: This deviates from the two other ways of ignoring properties.
Also, at least in v2 I guess `calcjob_ignored_attributes` can be done inside `node_ignored_attributes` as well, with entry point` "process.calculation.calcjob".

Maybe merging this is too complicated; but let's at least group the ignored properties under a field

```python
 'export_cache': {
            'default_cache_dir': str,
            'ignore': Schema({
                'calcjob_inputs':...,
                'calcjob_attributes':...,
                'node_attributes':...,
  }

```

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+
+#### utils
+
+
+def pytest_addoption(parser):
+    """Add pytest command line options."""
+    parser.addoption(
+        "--export-cache-allow-migration",
+        action="store_true",
+        default=False,
```
ltalirz, line 109: change?

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+        help="If True the stored archives can be migrated if needed."
+    )
+
+
+def _rehash_processes() -> None:
```
ltalirz, line 114: given that this module has grown to >600 lines of code, perhaps it would be worth moving a couple of the functions used here into a "utils" module (e.g. the next 3 functions below look like candidates)

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+    export_cache_config = testing_config.get('export_cache', {})
+
+    def _absolute_archive_path(filepath):
+        """
+        Returns the absolute filepath to the given archive.
+        The procedure is:
+
+        - If the path is already absolute, return it
+        - If the option default_cache_dir is given construct it relative to this
+        - Otherwise interpret the directory as relative to the test file inside a folder `caches`
```
ltalirz, line 196: this needs to be documented

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+                #Adapted from shared_datadir of pytest-datadir to not use paths
+                #in the tmp copies created by pytest
+                default_data_dir = pathlib.Path(request.fspath.dirname) / 'caches'
+            else:
+                default_data_dir = pathlib.Path(default_data_dir)
+            if not default_data_dir.exists():
+                default_data_dir.mkdir()
+
+            full_export_path = pathlib.Path(default_data_dir) / filepath
+            #print(full_export_path)
```
ltalirz, line 214: remove?

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+        """
+        default_data_dir = export_cache_config.get('default_data_dir', '')
+        filepath = pathlib.Path(filepath)
+
+        if filepath.is_absolute():
+            full_export_path = filepath
+        else:
+            if not default_data_dir:
+                #Adapted from shared_datadir of pytest-datadir to not use paths
+                #in the tmp copies created by pytest
```
ltalirz, line 206: this comment is not self-explanatory to me, could you add a bit more context (why not use paths, what are the tmp copies created by pytest?)

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+                default_data_dir = pathlib.Path(request.fspath.dirname) / 'caches'
+            else:
+                default_data_dir = pathlib.Path(default_data_dir)
+            if not default_data_dir.exists():
+                default_data_dir.mkdir()
```
ltalirz, line 211: this can fail.
you may try/except, with an exception message that explains how to adapt the configuration

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+
+            full_export_path = pathlib.Path(default_data_dir) / filepath
+            #print(full_export_path)
+        return os.fspath(full_export_path.absolute())
```
ltalirz, line 215: I think it makes sense to do the existence check here as well

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+        - If the path is already absolute, return it
+        - If the option default_cache_dir is given construct it relative to this
+        - Otherwise interpret the directory as relative to the test file inside a folder `caches`
+        """
+        default_data_dir = export_cache_config.get('default_data_dir', '')
+        filepath = pathlib.Path(filepath)
+
+        if filepath.is_absolute():
+            full_export_path = filepath
+        else:
```
ltalirz, line 203: what follows here is mostly about determining the data directory and could be moved into a separate private helper function that is independent of the file

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+            qub = QueryBuilder()
+            qub.append(queryclass, tag='node')  # query for CalcJobs nodes
+            to_export = [entry[0] for entry in qub.all()]
+            export_cache(node=to_export, savepath=full_export_path, overwrite=overwrite)
+
+    return _with_export_cache
+
+
+@pytest.fixture
+def hash_code_by_entrypoint(monkeypatch, testing_config):
```
ltalirz, line 362: The name of this fixture is not self-explanatory to me.

Reading the docstring, perhaps it could be called something like `liberal_hash`?

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+            full_export_path = pathlib.Path(default_data_dir) / filepath
+            #print(full_export_path)
+        return os.fspath(full_export_path.absolute())
+
+    return _absolute_archive_path
+
+
+@pytest.fixture(scope='function')
+def export_cache(hash_code_by_entrypoint, absolute_archive_path):
+    """Fixture to export an AiiDA graph from given node(s)"""
```
ltalirz, line 222: Could you include a usage example in the docstring?

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+            to_export = [node]
+
+        create_archive(
+            to_export, filename=full_export_path, overwrite=overwrite, include_comments=True
+        )  # extras are automatically included
+
+    return _export_cache
+
+
+# Do we always want to use hash_code_by_entrypoint here?
```
ltalirz, line 250: do we?

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+    """Fixture to load a cached AiiDA graph"""
+
+    def _load_cache(path_to_cache=None, node=None):
+        """
+        Function to import an AiiDA graph
+
+        :param path_to_cache: str or path to the AiiDA export file to load,
+                              if path_to_cache points to a directory,
+                              all import files in this dir are imported
+        :param node: AiiDA node, determines which cache to load (CURRENTLY NOT IMPLEMENTED)
```
ltalirz, line 262: do we need it then?
perhaps better to delete it and open a feature request

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+
+@pytest.fixture(scope='function')
+def absolute_archive_path(request, testing_config):
+    """
+    Fixture to get the absolute filepath for a given archive
+    """
+
+    export_cache_config = testing_config.get('export_cache', {})
+
+    def _absolute_archive_path(filepath):
```
ltalirz, line 189: I see it says `filepath` here, but it seems like the path can be a directory or a file?
If so, please rename this to `path` and document this.

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+        def _import_with_migrate(archive_path, *args, **kwargs):
+            #pylint: disable=import-outside-toplevel
+            from aiida.tools.archive import get_format
+            from aiida.common.exceptions import IncompatibleStorageSchema
+
+            try:
+                import_archive(archive_path, *args, **kwargs)
+            except IncompatibleStorageSchema:
+                if export_cache_allow_migration:
+                    echo(f'incompatible version detected for {archive_path}, trying migration')
```
ltalirz, line 48: I suggest you use `aiida.cmdline.utils.echo.echo_warning`

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+    and remove aiida-core version from hash
+    """
+    export_cache_config = testing_config.get('export_cache', {})
+
+    #Load the corresponding entry points
+    node_ignored_attributes = {
+        plugins.DataFactory(entry_point): ignored
+        for entry_point, ignored in export_cache_config.get('node_ignored_attributes', {})
+    }
+
```
ltalirz, line 375: it would be great if code duplication could be reduced a bit on the remaining code of this function

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+        # for this we rehash all process nodes
+        # this way we use the full caching mechanism of aiida-core.
+        # currently this should only cache CalcJobNodes
+        _rehash_processes()
+
+    return _load_cache
+
+
+@pytest.fixture(scope='function')
+def with_export_cache(export_cache, load_cache, absolute_archive_path):
```
ltalirz, line 308: Using this fixture would then be: `with with_export_cache(): ...`?

Having a look the tests you wrote, it seems to me that this context manager is actually the _recommended_ way of using the export cache functionality.


`with export_cache():` actually reads quite naturally, and the current `export_cache` fixture could be named `cache_node`?

I leave the question of naming up to you - if one were to create this functionality from scratch today, it would probably be called "archive cache" (`export` no longer exists as terminology in the verdi cli).
I might use the context manager as `with enable_archive_cache:` (or `with archive_cache:`) and add individual archives via `create_node_archive(...)`.

tests/export_cache/test_export_cache.py
```diff
@@ -0,0 +1,206 @@
...
+'''
+    assert node.is_finished_ok
+    assert res['computed_diff'].get_content() == res_diff
+
+    # now export cache
+    savepath = os.path.join(
+        os.path.dirname(os.path.abspath(__file__)), 'caches/diff_workchain.tar.gz'
+    )
+    if os.path.isfile(savepath):
+        os.remove(savepath)
```
ltalirz, line 80: Do users typically have to do this manually or can this be automated to some degree?

I guess at least the folder doesn't need to be specified (?)

Edit: I see, the "user friendly" fixture is actually `with_export_cache`.

tests/export_cache/test_export_cache.py
```diff
@@ -0,0 +1,206 @@
...
+    """test if mock of _get_objects_to_hash works for Code and Calcs"""
+
+    mock_code = mock_code_factory(
+        label='diff',
+        data_dir_abspath=os.path.join(os.path.dirname(os.path.abspath(__file__)), 'calc_data'),
+        entry_point=CALC_ENTRY_POINT,
+        ignore_paths=('_aiidasubmit.sh', 'file*')
+    )
+    objs = mock_code._get_objects_to_hash()
+    assert objs == [mock_code.get_attribute(key='input_plugin')]  #, mock_code.get_computer_name()]
```
ltalirz, line 114:
```suggestion
    assert objs == [mock_code.get_attribute(key='input_plugin')]
```

tests/export_cache/test_export_cache.py
```diff
@@ -0,0 +1,206 @@
...
+    Basic test of the run with cache fixture functionality,
+    should run workchain with cached calcjob
+    """
+    inputs = {'diff': generate_diff_inputs()}
+    #mock_code = mock_code_factory(
+    #    label='diff',
+    #    data_dir_abspath=os.path.join(os.path.dirname(os.path.abspath(__file__)), 'calc_data'),
+    #    entry_point=CALC_ENTRY_POINT,
+    #    ignore_paths=('_aiidasubmit.sh', 'file*')
+    #)
```
ltalirz, line 134:
```suggestion

```

tests/export_cache/test_export_cache.py
```diff
@@ -0,0 +1,206 @@
...
+    diff_code = aiida_local_code_factory(executable='diff', entry_point='diff')
+    diff_code.store()
+    inputs['diff']['code'] = diff_code
+    #builder = DiffWorkChain.get_builder()
+    data_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'caches')
+    #run_with_cache(builder=builder, data_dir=data_dir)
+    res, node = run_with_cache(builder=inputs, process_class=DiffWorkChain, data_dir=data_dir)
```
ltalirz, line 141: What is the difference between `run_with_cache` and running inside a `with_export_cache` context manager?
do we need both?

ltalirz (changes requested):
Thank you @janssenhenning  for contributing these improvements back upstream!

I have a few comments on the code, but they are all minor (feel free to take them or leave them) compared to one request: could you please add a few lines to the ["Using `export_cache`" section](https://aiida-testing.readthedocs.io/en/latest/user_guide/export_cache.html) of the documentation?

It doesn't need to be long; for comparison, here is the [corresponding section for `mock_code`](https://aiida-testing.readthedocs.io/en/latest/user_guide/mock_code.html).

tests/export_cache/test_export_cache.py
```diff
@@ -0,0 +1,206 @@
...
+    #mock_code = mock_code_factory(
+    #    label='diff',
+    #    data_dir_abspath=os.path.join(os.path.dirname(os.path.abspath(__file__)), 'calc_data'),
+    #    entry_point=CALC_ENTRY_POINT,
+    #    ignore_paths=('_aiidasubmit.sh', 'file*')
+    #)
+    diff_code = aiida_local_code_factory(executable='diff', entry_point='diff')
+    diff_code.store()
+    inputs['diff']['code'] = diff_code
+    data_dir = os.path.join(
```
ltalirz, line 184: that is not a directory, that is a file

aiida_testing/export_cache/_fixtures.py
```diff
@@ -0,0 +1,609 @@
...
+    Fixture to use in a with block
+    - Before the block the given cache is loaded (if it exists)
+    - within this block the caching of AiiDA is enabled.
+    - At the end an AiiDA export can be created (if test data should be overwritten)
+    Requires an absolute path to the export file to load or export to.
+    Export the provenance of all calcjobs nodes within the test.
+    """
+
+    @contextmanager
+    def _with_export_cache(data_dir_abspath, calculation_class=None, overwrite=False):
```
ltalirz, line 319: Why limit this to a single calculation class rather than a list?

Also, I would guess that, in practice, you users will typically anyhow want to target all calculation classes that were called as part of the workflow, am I wrong?

Would that make sense as the default behavior?

tests/export_cache/test_export_cache.py
janssenhenning: This is a remnant from the previous implementations of this feature. The `run_with_cache` fixture was the first to be implemented. The main difference between the two is that `run_with_cache` is limited to one process but will create a default archive name including the hash of the inputs  to the workchain. In `aiida-fleur` we by now only use `with_export_cache` but I think in aiida-kkr the `run_with_cache` fixture is still used. Maybe @PhilippRue can comment in more detail what the advantages are 

If not I agree that we should settle for only one

aiida_testing/export_cache/_fixtures.py
janssenhenning: I decided to go with the archive names, since I think it is more intuitive for people getting started with this plugins if they use the same terminology as the current AiiDA versions.

I took care of most of the suggestions you gave but I want to clean up the commit history a bit to make it easier to review, since the changes are larger
for the moment this is located here https://github.com/JuDFTteam/aiida-testing/tree/archive_cache and I will clean up the branch after the weekend

janssenhenning:
Superseded by #67

PR #57, opened 2022-07-19
Bugfix for specifying relative paths in .aiida-testing-config.yml

janssenhenning:
The path manipulation was using the full path of the config file
instead of the path to the directory of the config file

This was not noticed since the executable is only executed when
regenerating test data and no test was added in aiida-testing (My bad :()
A test is now added

### DIFF
--- a/aiida_testing/mock_code/_fixtures.py
+++ b/aiida_testing/mock_code/_fixtures.py
@@ -175,7 +175,7 @@ def mock_code_factory(
             mock_code_config[label] = code_executable_path
         if code_executable_path not in ('TO_SPECIFY', 'NOT_FOUND') and \
             not pathlib.Path(code_executable_path).is_absolute():
-            code_executable_path = os.fspath(_config.file_path / code_executable_path)
+            code_executable_path = os.fspath(_config.file_path.parent / code_executable_path)
 
         code = Code(
             input_plugin_name=entry_point,
--- a/aiida_testing/mock_code/_fixtures.py
+++ b/aiida_testing/mock_code/_fixtures.py
@@ -169,14 +169,29 @@ def mock_code_factory(
                 f"Configuration file {CONFIG_FILE_NAME} does not specify path to executable for code label '{label}'."
             )
         code_executable_path = mock_code_config.get(label, 'TO_SPECIFY')
-        if (not code_executable_path) and executable_name:
-            code_executable_path = shutil.which(executable_name) or 'NOT_FOUND'
+
+        if code_executable_path != 'TO_SPECIFY' and \
+           not pathlib.Path(code_executable_path).is_absolute():
+            #Relative paths are interpreted with respect to the
+            #aiida-testing-config.yml file
+            relative_path = _config.file_path.parent / code_executable_path
+            code_executable_path = os.fspath(relative_path)
+            if not relative_path.exists():
+                raise ValueError(
+                    f"Relative path {code_executable_path} in {CONFIG_FILE_NAME} "
+                    f"does not exist for code label '{label}'."
+                )
+
+        elif code_executable_path == 'TO_SPECIFY' and executable_name:
+            code_executable_path = shutil.which(executable_name)
+            if code_executable_path is None:
+                raise ValueError(
+                    f"Executable {executable_name} not found on PATH for code label '{label}'."
+                )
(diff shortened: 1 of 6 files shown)


aiida_testing/mock_code/_fixtures.py
```diff
@@ -175,7 +175,7 @@ def _get_mock_code(
             mock_code_config[label] = code_executable_path
         if code_executable_path not in ('TO_SPECIFY', 'NOT_FOUND') and \
             not pathlib.Path(code_executable_path).is_absolute():
-            code_executable_path = os.fspath(_config.file_path / code_executable_path)
+            code_executable_path = os.fspath(_config.file_path.parent / code_executable_path)
```
ltalirz: Just wondering what should be the logical order of these different approaches (also on the lines above).

Here is what would make sense to me:

 1. check if it is an absolute path. if yes, take that
 2. check if it is a relative path with respect to the config file. if yes, take that
 3. check if the executable is in the PATH

Would you agree?

ltalirz (commented):
thanks for the fix @janssenhenning , 
I just have one small suggestion about the order of how things are checked

aiida_testing/mock_code/_fixtures.py
janssenhenning: Yes I think that is a sensible order 👍 
I have one question about this code part
It starts with 
```python
code_executable_path = mock_code_config.get(label, 'TO_SPECIFY')
if (not code_executable_path) and executable_name:
    code_executable_path = shutil.which(executable_name) or 'NOT_FOUND'
```
If the code is not specified at all it will never take the `executable_name` into account because of the default
So should the executable_name only be relevant if you set the code to an empty string or is this a mistake?

ltalirz: Hey @janssenhenning , sorry for the delay.

Indeed, that is a bug and it turns out it was introduced by myself - shame on me, also for not adding any test for the `executable_name` feature ;-)

I believe my idea would have been that you could forego the need for the config file simply by passing the name of the executable to the pytest fixture and making sure it was in the PATH when the tests are run.

As you point out, at least as the code stands now that would actually only work when using a config file that explicitly sets the code path to an empty value.

I can't remember ever purposefully doing that - given this, I would also be ok with removing the `executable_name` parameter altogether (I don't think it will break anything).
I leave it up to you whether you want to make it work or remove it.

ltalirz: Just waiting on reply from @janssenhenning before moving forward

ltalirz: ping

janssenhenning: Sorry for being inactive on this. I now adjusted the path determination to this order and added a test for the executable name given case.

I also added two more error messages for the cases, where
- A relative path is given but no executable is found at that path
- no path but the executable name is given and it is not found on the PATH

janssenhenning:
@ltalirz I don't know what to make of the failing tests. Did something change in how the AiiDA test profile has to be set up?

ltalirz:
Thanks @janssenhenning ! I'll have a look into the erorr

ltalirz:
Fixed, can you please rebase on `develop` ?

tests/mock_code/test_diff.py
```diff
@@ -172,3 +186,59 @@ def test_regenerate_test_data(mock_code_factory, generate_diff_inputs, datadir):
     # check that ignore_paths works
     assert not (datadir / '_aiidasubmit.sh').is_file()
     assert (datadir / 'file1.txt').is_file()
+
+
```
ltalirz, line 190: Instead of duplicating the `regenerate_test_data` code, one could use `@pytest.mark.parametrize` here.

I would guess something like
```python
@pytest.mark.parametrize("kwargs", [{'label': 'diff-relative'}, {...}])
def test_regenerate_test_data_relative(kwargs, mock_code_factory, generate_diff_inputs, datadir)
  mock_code = mock_code_factory(
      **kwargs,
      data_dir_abspath=...
```

ltalirz (approved):
Great, thank you @janssenhenning !

I guess one could still reduce code duplication in the tests a bit, but that could also be done in a later PR.

PR #61, opened 2022-11-21
🔧 Modernise build tools

chrisjsewell:
Use flit and move all the tool configuration into pyproject.toml


pyproject.toml
```diff
@@ -1,6 +1,80 @@
...
+    "pylint~=2.12.2",
+    "mypy==0.930",
+]
+
+[project.scripts]
+aiida-mock-code = "aiida_testing.mock_code._cli:run"
+
+[project.entry-points."pytest11"]
+aiida_mock_code = "aiida_testing.mock_code"
+aiida_export_cache = "aiida_testing.export_cache"
```
chrisjsewell, line 77: FYI you don't need to load via `pytest_plugins`, if you include this entry point 😉

ltalirz: Oh, nice feature!

Also a bit "dangerous" - I guess any package can now drop any fixture into your pytest setup ;-)

ltalirz (approved):
thanks a lot @chrisjsewell , all good!

pyproject.toml
chrisjsewell: well you likely wouldn't be installing this package if you didn't want to use the fixtures 😄

PR #63, opened 2022-11-22
👌 IMPROVE failure logging 

chrisjsewell:
Log run information from mock code execution, and print to stdout on test fixture teardown.

For example:

```
        check_diff_output(res)
    
>       raise
E       RuntimeError: No active exception to reraise

tests/mock_code/test_diff.py:98: RuntimeError
------------------------------------------- Captured log setup -------------------------------------------
DEBUG    pgsu:__init__.py:143 Trying to connect via "psycopg2"...
---------------------------------------- Captured stdout teardown ----------------------------------------
AiiDA mock code logging:
2022-11-22 07:23:37.801306:diff: Init mock code
2022-11-22 07:23:37.802971:diff: Hashed paths: ['_aiidasubmit.sh', '_scheduler-stderr.txt', '_scheduler-stdout.txt', 'file1.txt', 'file2.txt', 'patch.diff']
2022-11-22 07:23:37.803233:diff: Cached directory exists: /Users/chrisjsewell/Documents/GitHub/aiida-testing/tests/mock_code/data/mock-diff-4b5ca93b23993979a24f795147487f1c

======================================== short test summary info =========================================
FAILED tests/mock_code/test_diff.py::test_basic - RuntimeError: No active exception to reraise
```

Also adds `--mock-fail-on-missing` CLI option.
This can be used for CI (e.g. GH Action) runs, where you should likely not be wanting to generate new cache data

### DIFF
--- a/aiida_testing/mock_code/__init__.py
+++ b/aiida_testing/mock_code/__init__.py
@@ -14,3 +14,6 @@ __all__ = (
     "testing_config",
     "mock_code_factory",
 )
+
+# ensure aiida's pytest plugin is loaded, which we rely on
+pytest_plugins = ['aiida.manage.tests.pytest_fixtures']
--- a/aiida_testing/mock_code/_cli.py
+++ b/aiida_testing/mock_code/_cli.py
@@ -4,6 +4,7 @@
 Implements the executable for running a mock AiiDA code.
 """
 
+from datetime import datetime
 import os
 import sys
 import shutil
@@ -27,23 +28,44 @@ def run() -> None:
     directory.
     """
     # Get environment variables
+    log_path = Path(os.environ[EnvKeys.LOG_FILE.value])
     label = os.environ[EnvKeys.LABEL.value]
     data_dir = os.environ[EnvKeys.DATA_DIR.value]
     executable_path = os.environ[EnvKeys.EXECUTABLE_PATH.value]
     ignore_files = os.environ[EnvKeys.IGNORE_FILES.value].split(':')
     ignore_paths = os.environ[EnvKeys.IGNORE_PATHS.value].split(':')
     regenerate_data = os.environ[EnvKeys.REGENERATE_DATA.value] == 'True'
+    fail_on_missing = os.environ[EnvKeys.FAIL_ON_MISSING.value] == 'True'
 
-    hash_digest = get_hash().hexdigest()
+    def _log(msg: str) -> None:
+        """Write a message to the log file."""
+        with open(log_path, 'a', encoding='utf8') as log_file:
+            log_file.write(f"{datetime.now()}:{label}: {msg}\n")
+
+    _log('Init mock code')
+
+    hash_digest = get_hash(_log).hexdigest()
 
     res_dir = Path(data_dir) / f"mock-{label}-{hash_digest}"
 
-    if regenerate_data and res_dir.exists():
-        shutil.rmtree(res_dir)
+    if res_dir.exists():
+        _log(f"Cached directory exists: {res_dir}")
+        if regenerate_data:
+            _log("Regenerating data")
+            shutil.rmtree(res_dir)
+    else:
+        _log(f"Cached directory does not exist: {res_dir}")
+
+        if fail_on_missing:
+            _log(f"ERROR: Failing on missing cached data: {res_dir}")
+            sys.exit("Missing cached data")
 
     if not res_dir.exists():
         if not executable_path:
-            sys.exit("No existing output, and no executable specified.")
+            _log(f"Failing on missing cached data: {res_dir}")
+            sys.exit("ERROR: No existing output, and no executable specified.")
+
+        _log(f"Regenerating with executable: {executable_path}")
 
         subprocess.call([executable_path, *sys.argv[1:]])
 
@@ -65,16 +87,18 @@ def run() -> None:
             elif path.is_file():
                 shutil.copyfile(path, path.name)
             else:
+                _log(f"ERROR: Can not copy '{path.name}'.")
                 sys.exit(f"Can not copy '{path.name}'.")
 
 
-def get_hash() -> 'hashlib._Hash':
+def get_hash(logger: ty.Callable[[str], None]) -> 'hashlib._Hash':
     """
     Get the MD5 hash for the current working directory.
     """
     md5sum = hashlib.md5()
     # Here the order needs to be consistent, thus globbing
     # with 'sorted'.
+    used_paths = []
     for path in sorted(Path('.').glob('**/*')):
         if path.is_file() and not path.match('.aiida/**'):
             with open(path, 'rb') as file_obj:
@@ -83,6 +107,9 @@ def get_hash() -> 'hashlib._Hash':
                 file_content_bytes = strip_submit_content(file_content_bytes)
             md5sum.update(path.name.encode())
             md5sum.update(file_content_bytes)
+            used_paths.append(str(path))
+
+    logger(f"Hashed paths: {used_paths}")
 
     return md5sum
 
--- a/aiida_testing/mock_code/_cli.py
+++ b/aiida_testing/mock_code/_cli.py
@@ -64,7 +64,7 @@ def run() -> None:
             _log("ERROR: No existing cache, and no executable specified.")
             sys.exit("No existing cache, and no executable specified.")
 
-        _log(f"Regenerating with executable: {executable_path}")
+        _log(f"Running with executable: {executable_path}")
 
         subprocess.call([executable_path, *sys.argv[1:]])
 
--- a/aiida_testing/mock_code/_cli.py
+++ b/aiida_testing/mock_code/_cli.py
@@ -53,17 +53,16 @@ def run() -> None:
         if regenerate_data:
             _log("Regenerating data")
             shutil.rmtree(res_dir)
+    elif fail_on_missing:
+        _log(f"ERROR: Missing cached data: {res_dir}")
+        sys.exit("Missing cached data")
     else:
         _log(f"Cached directory does not exist: {res_dir}")
 
-        if fail_on_missing:
-            _log(f"ERROR: Failing on missing cached data: {res_dir}")
-            sys.exit("Missing cached data")
-
     if not res_dir.exists():
         if not executable_path:
-            _log(f"Failing on missing cached data: {res_dir}")
-            sys.exit("ERROR: No existing output, and no executable specified.")
+            _log("ERROR: No existing cache, and no executable specified.")
+            sys.exit("No existing cache, and no executable specified.")
 
         _log(f"Regenerating with executable: {executable_path}")
 
--- a/aiida_testing/mock_code/_cli.py
+++ b/aiida_testing/mock_code/_cli.py
@@ -49,15 +49,15 @@ def run() -> None:
     res_dir = Path(data_dir) / f"mock-{label}-{hash_digest}"
 
     if res_dir.exists():
-        _log(f"Cached directory exists: {res_dir}")
+        _log(f"Cache hit: {res_dir}")
         if regenerate_data:
             _log("Regenerating data")
             shutil.rmtree(res_dir)
     elif fail_on_missing:
-        _log(f"ERROR: Missing cached data: {res_dir}")
+        _log(f"ERROR: No cache hit for: {res_dir}")
         sys.exit("Missing cached data")
     else:
-        _log(f"Cached directory does not exist: {res_dir}")
+        _log(f"No cache hit for: {res_dir}")
 
     if not res_dir.exists():
         if not executable_path:
--- a/aiida_testing/mock_code/_fixtures.py
+++ b/aiida_testing/mock_code/_fixtures.py
@@ -191,29 +191,25 @@ def mock_code_factory(
             raise ValueError(
                 f"Configuration file {CONFIG_FILE_NAME} does not specify path to executable for code label '{label}'."
             )
-        code_executable_path = mock_code_config.get(label, 'TO_SPECIFY')
 
-        if code_executable_path != 'TO_SPECIFY' and \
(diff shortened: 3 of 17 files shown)


aiida_testing/mock_code/_cli.py
```diff
@@ -27,23 +28,44 @@ def run() -> None:
...
     executable_path = os.environ[EnvKeys.EXECUTABLE_PATH.value]
     ignore_files = os.environ[EnvKeys.IGNORE_FILES.value].split(':')
     ignore_paths = os.environ[EnvKeys.IGNORE_PATHS.value].split(':')
     regenerate_data = os.environ[EnvKeys.REGENERATE_DATA.value] == 'True'
+    fail_on_missing = os.environ[EnvKeys.FAIL_ON_MISSING.value] == 'True'
 
-    hash_digest = get_hash().hexdigest()
+    def _log(msg: str) -> None:
+        """Write a message to the log file."""
+        with open(log_path, 'a', encoding='utf8') as log_file:
```
chrisjsewell, line 42: I think this should be ok, for potentially concurrent writes to the log file?
Or do you write a separate log file, for every mock code execution?

aiida_testing/mock_code/_cli.py
```diff
@@ -27,23 +28,44 @@ def run() -> None:
...
+        _log(f"Cached directory exists: {res_dir}")
+        if regenerate_data:
+            _log("Regenerating data")
+            shutil.rmtree(res_dir)
+    else:
+        _log(f"Cached directory does not exist: {res_dir}")
+
+        if fail_on_missing:
+            _log(f"ERROR: Failing on missing cached data: {res_dir}")
+            sys.exit("Missing cached data")
```
chrisjsewell: I also want to add the ability here, to dump the current folder somewhere, so that you can look through the files and e.g. look at diffs between existing cached folders, to understand why they are creating different hashes (e.g. for the case with the `.xyz` files being different due to rounding errors)

aiida_testing/mock_code/_cli.py
```diff
@@ -27,23 +28,44 @@ def run() -> None:
...
 
     if not res_dir.exists():
         if not executable_path:
-            sys.exit("No existing output, and no executable specified.")
+            _log(f"Failing on missing cached data: {res_dir}")
```
ltalirz:
```suggestion
            _log(f"ERROR: Failing on missing cached data: {res_dir}")
```

aiida_testing/mock_code/_cli.py
```diff
@@ -27,23 +28,44 @@ def run() -> None:
...
+            sys.exit("ERROR: No existing output, and no executable specified.")
```
ltalirz:
```suggestion
            sys.exit("No existing output, and no executable specified.")
```

aiida_testing/mock_code/_fixtures.py
```diff
@@ -189,9 +206,11 @@ def _get_mock_code(
                     f"Executable {executable_name} not found on PATH for code label '{label}'."
                 )
 
+        if code_executable_path == 'TO_SPECIFY':
+            code_executable_path = ""
```
ltalirz: I guess in that case, the `TO_SPECIFY` string can be removed entirely from the code above

aiida_testing/mock_code/_cli.py
ltalirz: I think this solution is fine.
It should indeed allow concurrent writes, and in the end this is about debugging; people will be able to read this info - let's not make it more complicated than needed

ltalirz (commented):
this is great, thanks a lot @chrisjsewell !

should we add perhaps one or two small tests, one checking the content of the log file, and one checking that execution fails in the expected way when asking to fail on missing executable?

chrisjsewell:
> should we add perhaps one or two small tests, one checking the content of the log file, and one checking that execution fails in the expected way when asking to fail on missing executable?

yep absolutely, just wanted to open this prematurely, so you knew I was working on it 😄

aiida_testing/mock_code/_fixtures.py
chrisjsewell: yeh I just know this issue come up with aiida-lsmo, that when we didn't have the `.aiida-testing-config.yml` present it was trying to run `TO_SPECIFY` as if that was the executable path 😅

aiida_testing/mock_code/_cli.py
ltalirz: That makes sense to me as well - although for the use case of comparing to results on CI, the user would need to [add an artifact](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#example) to their github actions build, i.e. it's not quite as straightforward as the logs in the test output.

We could document how to do this, of course.

aiida_testing/mock_code/_cli.py
chrisjsewell: ah, this was meant to be in the log message not the exit one

chrisjsewell:
FYI @ltalirz the other thing I was thinking to do, is allow users to specify a function as a string, to be loaded by https://docs.python.org/3/library/importlib.html#importlib.import_module in the mock code (or perhaps using entry points), that could override the default hashing function,
e.g. to deal with things like the rounding errors in `.xyz` files

ltalirz:
> FYI @ltalirz the other thing I was thinking to do, is allow users to specify a function as a string, to be loaded by https://docs.python.org/3/library/importlib.html#importlib.import_module in the mock code (or perhaps using entry points), that could override the default hashing function, e.g. to deal with things like the rounding errors in `.xyz` files

`mock_code` has been in use for a while and the issue raised by @mpougin is the first time this has come up to my knowledge.

Yesterday, she posted the following comparison between two XYZ files 
https://github.com/lsmo-epfl/aiida-lsmo/issues/102#issuecomment-1322512168
which does not look like a rounding error at all.

If we do determine that this is not a bug but indeed related to randomness at the level of the machine epsilon, then I agree we can implement such a functionality (although I suspect its use will be limited to "experts").

aiida_testing/mock_code/__init__.py
```diff
@@ -14,3 +14,6 @@
     "testing_config",
     "mock_code_factory",
 )
+
+# ensure aiida's pytest plugin is loaded, which we rely on
+pytest_plugins = ['aiida.manage.tests.pytest_fixtures']
```
chrisjsewell, line 19: @ltalirz now users don't have to use `pytest_plugins` at all

aiida_testing/mock_code/_fixtures.py
```diff
@@ -145,6 +162,12 @@ def _get_mock_code(
             assert isinstance(arg, collections.abc.Iterable) and not isinstance(arg, str), \
                 f"'ignore_files' and 'ignore_paths' arguments must be tuples or lists, found {type(arg)}"
 
+        if entry_point is None:
+            entry_point = label
+        if data_dir_abspath is None:
+            request.node.path.parent.joinpath("data").mkdir(exist_ok=True)
+            data_dir_abspath = request.node.path.parent / "data"
```
chrisjsewell, line 169: @ltalirz I added the ability to just specify the label, e.g. `mock_code_factory('diff')`, in which case the entry point will be the same as the label, and the data directory is set as a `data` folder in the test's parent directory

aiida_testing/mock_code/_fixtures.py
chrisjsewell: done

aiida_testing/mock_code/_cli.py
chrisjsewell: I'll do this dumping in another PR

aiida_testing/mock_code/__init__.py
ltalirz: Hm... so defining the `pytest_plugins` variable in any part of the code will be picked up by `pytest`?

I read [here](https://docs.pytest.org/en/latest/reference/reference.html#globalvar-pytest_plugins) that it should be defined in test modules or the conftest file.

I would personally vote to let the users explicitly specify this, so that they are at least aware of which files the fixtures are read from. For my taste the fixtures are already "magic enough".

chrisjsewell: but you literally cannot use the mock code fixtures without it, its essentially a dependency of the plugin

chrisjsewell: IMO enforcing that every user has to specify `pytest_plugins` is just adding extra complication

aiida_testing/mock_code/_cli.py
```diff
@@ -27,23 +28,43 @@ def run() -> None:
...
+    _log('Init mock code')
+
+    hash_digest = get_hash(_log).hexdigest()
 
     res_dir = Path(data_dir) / f"mock-{label}-{hash_digest}"
 
-    if regenerate_data and res_dir.exists():
-        shutil.rmtree(res_dir)
+    if res_dir.exists():
+        _log(f"Cached directory exists: {res_dir}")
```
ltalirz:
```suggestion
        _log(f"Cache hit: {res_dir}")
```

aiida_testing/mock_code/_cli.py
```diff
@@ -27,23 +28,43 @@ def run() -> None:
...
+        if regenerate_data:
+            _log("Regenerating data")
+            shutil.rmtree(res_dir)
+    elif fail_on_missing:
+        _log(f"ERROR: Missing cached data: {res_dir}")
```
ltalirz:
```suggestion
        _log(f"ERROR: No cache hit for: {res_dir}")
```

aiida_testing/mock_code/_cli.py
```diff
@@ -27,23 +28,43 @@ def run() -> None:
...
+        sys.exit("Missing cached data")
+    else:
+        _log(f"Cached directory does not exist: {res_dir}")
```
ltalirz:
```suggestion
        _log(f"No cache hit for: {res_dir}")
```

aiida_testing/mock_code/_fixtures.py
ltalirz: Thank you, these are good choices for defaults

tests/mock_code/test_failures.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def test_failure_no_cache(pytester: pytest.Pytester):
+    """Test when no cache is found and --mock-fail-on-missing is set."""
+    pytester.makeconftest(CONFTEST)
+    pytester.path.joinpath("file1.txt").write_text("a")
+    pytester.path.joinpath("file2.txt").write_text("b")
+    pytester.makepyfile(PYFILE)
+    result = pytester.runpytest_subprocess("-k", "test_basic", "--mock-fail-on-missing")
+    result.stdout.re_match_lines([r".*ERROR: Missing cached data.*"])
```
ltalirz:
```suggestion
    result.stdout.re_match_lines([r".*ERROR: No cache hit for.*"])
```

ltalirz (dismissed):
Cheers, the remaining changes look fine; just some minor suggestions (optional). Let me know whether you want to add them

aiida_testing/mock_code/__init__.py
ltalirz: Yeah, with the magic insertion of the fixtures via the `pyproject.toml` it feels like an unnecessary burden (obviously that magic insertion is equivalent).

But I fear a bit that users will no longer know which fixtures are available in their tests, and where to look for them in the code...

Also, about the `pyproject.toml`, I'm wondering: I may want to run the test suite of other packages in my aiida environment - say, `pgsu`.
Since the `aiida_profile` fixture is `autouse=True`, does this then mean that it will be used in the test suite of my other non-AiiDA package as well (where it will just slow down the test suite)?

chrisjsewell: > Since the `aiida_profile fixture` is `autouse=True`,

Ha well I never quite got why this needed to be the case in the first place 😅
I believe you made it autouse in: https://github.com/aiidateam/aiida-core/pull/3319, is there reason why it has to be?

chrisjsewell: > I may want to run the test suite of other packages in my aiida environment

IMO thats a very esoteric use case though; I use tox and thus have a different virtual environment created for every different test suite I run, I feel thats a better use of environments

> But I fear a bit that users will no longer know which fixtures are available in their tests, and where to look for them in the code...

you just run `pytest --fixtures` (as I've already put in the documentation here)

chrisjsewell:
> Cheers, the remaining changes look fine; just some minor suggestions (optional). Let me know whether you want to add them

added cheers

ltalirz (approved):
thank you @chrisjsewell !

chrisjsewell:
cheers

aiida_testing/mock_code/__init__.py
ltalirz: Just got bitten by this...

@janssenhenning I'd be curious about your input on this: do you prefer if the `aiida-testing` fixtures are made available automatically via `pip install aiida-testing` or do you prefer to explicitly add then to the `conftest.py` via `pytest_plugins = ["aiida_testing.mock_code"]` ?

janssenhenning: I don't mind either way, probably easier to have them available automatically. One drawback is that people not very familiar with the test systems will maybe need more time to find out where fixtures come from. But the difference here between looking into the `conftest.py` or `pyproject.toml` for the test system dependencies is minimal.

But, do you know whether there is something like `pytest.importorskip` to skip tests when a fixtures are not available?

ltalirz: Thanks for the feedback. In that case, let's keep the current implementation for the moment.

> But, do you know whether there is something like `pytest.importorskip` to skip tests when a fixtures are not available?

I don't know how to skip a test based on the availability of a fixture

PR #65, opened 2022-11-23
✨ NEW: Allow for custom hasher to mock_code_factory

chrisjsewell:


tests/mock_code/test_isolated.py
```diff
@@ -61,3 +61,29 @@ def test_failure_no_executable(pytester: pytest.Pytester):
...
+        from aiida.engine import run_get_node
+        from aiida_testing.mock_code import InputHasher
+        class CustomHasher(InputHasher):
+            def modify_content(self, path, content):
+                if path.name == 'file1.txt':
+                    self.log("Skipping file1.txt")
+                    return None
+                return content
+        def test_basic(mock_code_factory, generate_diff_inputs):
+            mock_code = mock_code_factory('diff', hasher=CustomHasher)
```
chrisjsewell, line 82: Primarily, you can see the use here

aiida_testing/mock_code/_env_keys.py
```diff
@@ -2,20 +2,96 @@
...
+            return load_hasher(file_path, class_name)
+        return self._hasher
+
+    def to_env(self) -> str:
+        """
+        Return a string that can be used to export the environmental variables
+        """
+        string = inspect.cleandoc(
+            f"""
+                export {_EnvKeys.LOG_FILE.value}="{self.log_file}"
```
ltalirz, line 64: I guess people won't really look at the generated file but that's quite a bit of indentation here...

tests/mock_code/test_isolated.py
```diff
@@ -61,3 +61,29 @@ def test_failure_no_executable(pytester: pytest.Pytester):
...
+            builder = mock_code.get_builder()
+            run_get_node(builder, **generate_diff_inputs())
+            assert False
```
ltalirz, line 85: What is this line doing here?

ltalirz (approved):
thanks @chrisjsewell !

happy to add the feature; although I think the goal of this package should be that most users never need to resort to it.

Just two minor comments; feel free to merge afterwards, no need to re-review

aiida_testing/mock_code/_env_keys.py
chrisjsewell: Oh the cleandoc function it is wrapped in removes all that indentation

greschd: Interesting that the stdlib has two functions that do almost the same.. I've always used [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) for this. 🙂

Guess `cleandoc` is nicer because you don't need to care about the first line being different.

tests/mock_code/test_isolated.py
chrisjsewell: Making this test fail, so that pytest outputs the full "log" for it, in this case primarily tue captured stdout with the mock code log in it. If the test fails you don't see this.

aiida_testing/mock_code/_env_keys.py
chrisjsewell: Yep https://docs.python.org/3/library/inspect.html#inspect.cleandoc is similar

PR #66, opened 2022-11-24
👌 Parse test name to mock executable

chrisjsewell:
Can be useful for the hash hook (and if we want to add a hook to dump inputs with no cache-hit)


ltalirz (approved):
thanks @chrisjsewell , that looks useful.

I guess you are also planning of using it?

chrisjsewell:
> I guess you are also planning of using it?

So in terms of the hashing: I was thinking to use it for https://github.com/lsmo-epfl/aiida-lsmo/pull/108; to read in and manipulate the `.xyz` file, only for the specific test that was failing. But for now I've just gone for a more simple fix.

For outputting input files for failed cache hits (not yet implemented), it would certainly be useful though; to put them in an output folder based on the test name.

PR #67, opened 2023-01-05
Archive cache fixtures

janssenhenning:
Supersedes #56 (and thus also #29)

Sorry about the new PR, but I moved aiida-fleur to use the branch of #56 and did not want to rename the fixture names there until aiida-testing has a stable release of these fixtures. (AFAIK you can't change the source branch of a PR, right?)

The names of the fixtures were changed to more accurately reflect their purpose and agree with the current terminology in aiida-core

- `enable_archive_cache`: previously `with_export_cache`
- `create/load_node_archive`: previously `export_cache/load_cache`
- `liberal_hash`: `hash_node_by_entrypoint`

Flipped the logic of archive migration. Now done by default and only when the flag `--archive-cache-forbid-migration` is given it is not migrated. Additionally migration is done in a temporary directory to not modify the test files if not asked for explicitly

Removed `run_with_cache` fixture as it is the same as `enable_archive_cache` with only slightly different usage
Added usage guide


ltalirz:
thanks @janssenhenning and sorry for the delay on my side (very busy).

is there any part part of the code you'd like me to look at in particular?

are any of the comments from https://github.com/aiidateam/aiida-testing/pull/56 left unaddressed?

janssenhenning:
> thanks @janssenhenning and sorry for the delay on my side (very busy).
> 
@ltalirz No problem

> is there any part part of the code you'd like me to look at in particular?
> 
> are any of the comments from #56 left unaddressed?

Probably the most important part is, whether the logic of determining the filepaths is now understandable and well documented
I also made one additional imrpovement here. When a archive already exists and during a test run it has to be migrated to the correct version, this is now done in a temporary directory to remove unnecessary noise in the git repository (if the `overwrite=True` option is not passed to `enable_archive_cache`)
Would it make sense to add another commandline flag for the overwrite option to overwrite all archive caches?

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,109 @@
...
+
+        inputs = {
+            'diff': {
+                'code': diff_code,
+                'parameters': parameters,
+                'file1': file1,
+                'file2': file2,
+            }
+        }
+    with enable_archive_cache('diff_workchain.aiida'):
```
ltalirz: indent, please

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,109 @@
+===========================
+Using :mod:`.archive_cache`
+===========================
+
+:mod:`.archive_cache` provides three central fixtures 
+
+- :py:func:`~aiida_testing.archive_cache.create_node_archive` and :py:func:`~aiida_testing.archive_cache.load_node_archive` are used to manage AiiDA archive test files in the pytest environment
```
ltalirz: Do we still need these?

Given that they are not documented below, I suggest you either move the mention of them to the end of the document or you remove them (fixtures + documentation) entirely.

It appears new users can really focus just on the `enable_archive_cache` fixture.

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,109 @@
...
+        }
+    with enable_archive_cache('diff_workchain.aiida'):
+        res, node = run_get_node(DiffWorkChain, **inputs)
+
+    # Test results of workchain
+
+The fixture will look for the AiiDA archive named ``diff_workchain.aiida`` in a folder named ``caches`` in the same directory as the test file, if nothing else is specified.
+If this exists the archive is imported and the AiiDA caching functionality is enabled. All calculations created inside the with block will use the cached nodes if their
+inputs and attributes match.
+If the archive does not exist the workchain will try to run the complete calculation.
```
ltalirz:
```suggestion
If the archive does not exist the workchain will try to run the complete calculation and, afterwards, create the archive in the specified location.
```
am I right?

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,109 @@
...
+
+.. note::
+    The caching mechanism of AiiDA is modified within test functions using these fixtures to ignore attributes, that would break the
+    caching if the tests are run on different machines with different versions of AiiDA for example
+    By default the stored test caches are also migrated to match the installed AiiDA version
```
ltalirz:
```suggestion
    The hashing mechanism of AiiDA is modified within tests that use the fixture to ignore certain attributes that would invalidate the
    cache when running the tests on different machines, with different versions of AiiDA, etc.
    By default, the test archives are migrated to match the installed AiiDA version.
```

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,109 @@
...
+
+
+The following options can be specified in the ``aiida-testing-config.yml`` file
+
+.. code-block:: yaml
+
+    archive_cache:
+        default_data_dir: ... #If specified all relative paths passed to enable_archive_cache are relative to this
+        ignore:
+            calcjob_inputs: [...] #List of link labels of inputs to ignore in the aiida hash
```
ltalirz, line 81: can you give an example in the docs of when you may need to modify this?

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,109 @@
...
+inputs and attributes match.
+If the archive does not exist the workchain will try to run the complete calculation.
+
+.. note::
+    The caching mechanism of AiiDA is modified within test functions using these fixtures to ignore attributes, that would break the
+    caching if the tests are run on different machines with different versions of AiiDA for example
+    By default the stored test caches are also migrated to match the installed AiiDA version
+
+
+The following options can be specified in the ``aiida-testing-config.yml`` file
```
ltalirz: Maybe mention here that none of these options need to be modified/set in order to use `archive_cache` (I guess it will work just out of the box?).

docs/source/user_guide/archive_cache.rst
ltalirz: I guess this means you are ignoring the Computer on which the Code is set up.
Are you including any information about the Code in the hash?

ltalirz (commented):
thanks a lot @janssenhenning ; just went through the new docs!

ltalirz:
> Would it make sense to add another commandline flag for the overwrite option to overwrite all archive caches?

I think so, yes

ltalirz:
> Probably the most important part is, whether the logic of determining the filepaths is now understandable and well documented

That came out quite clearly in the docs, thanks!

docs/source/user_guide/archive_cache.rst
janssenhenning: Only the name of the calculation plugin is included in the hash of the code

docs/source/user_guide/archive_cache.rst
janssenhenning: transformed these into normal functions and moved them into the utils module 👍

docs/source/user_guide/archive_cache.rst
janssenhenning: yes

docs/source/user_guide/archive_cache.rst
janssenhenning: I wrote one paragraph giving the example of testing across AiiDA 1.X to 2.X

janssenhenning:
@ltalirz addressed all your comments and added `--archive-cache-overwrite` option to replace all archives in a test run

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,117 @@
+===========================
+Using :mod:`.archive_cache`
+===========================
+
+:mod:`.archive_cache` provides three central fixture
+
+- :py:func:`~aiida_testing.archive_cache.enable_archive_cache` uses the AiiDA caching mechanisms to enable end-to-end tests of high-level workchains without executing individual `Calcjob` processes, if they can be cached from a stored archive
```
ltalirz:
```suggestion
:mod:`.archive_cache` provides the :py:func:`~aiida_testing.archive_cache.enable_archive_cache` pytest fixture. It uses the AiiDA caching mechanisms to enable end-to-end tests of high-level workchains without executing individual `Calcjob` processes, if they can be cached from a stored archive.
```

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Here an example for using the :py:func:`~aiida_testing.archive_cache.enable_archive_cache` to test a simple workchain using the ``diff`` code
```
ltalirz:
```suggestion
Here an example of using the :py:func:`~aiida_testing.archive_cache.enable_archive_cache` fixture to test a simple workchain using the ``diff`` code
```

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,117 @@
...
+        running = self.submit(DiffCalculation, **inputs)
+
+        return ToContext(diff_calc=running)
+
+    def results(self):
+        computed_diff = self.ctx.diff_calc.get_outgoing().get_node_by_label('diff')
+        self.out('computed_diff', computed_diff)
+
+
+And an example of using the above fixtures in a pytest test suite looks as follows
```
ltalirz:
```suggestion
And here is the test
```

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,117 @@
...
+        }
+        with enable_archive_cache('diff_workchain.aiida'):
+            res, node = run_get_node(DiffWorkChain, **inputs)
+
+        # Test results of workchain
+
+The fixture will look for the AiiDA archive named ``diff_workchain.aiida`` in a folder named ``caches`` in the same directory as the test file, if nothing else is specified.
+If this exists the archive is imported and the AiiDA caching functionality is enabled. All calculations created inside the with block will use the cached nodes if their
+inputs and attributes match.
+If the archive does not exist the workchain will try to run the complete calculation and, afterwards, create the archive in the specified location.
```
ltalirz:
```suggestion
If the archive does not exist the, workchain will try to run the complete calculation and, afterwards, create the archive in the specified location.
```

docs/source/user_guide/archive_cache.rst
```diff
@@ -0,0 +1,117 @@
...
+    archive_cache:
+        default_data_dir: ... #If specified all relative paths passed to enable_archive_cache are relative to this
+        ignore:
+            calcjob_inputs: [...] #List of link labels of inputs to ignore in the aiida hash
+            calcjob_attributes: [...] #List of attributes of CalcjobNodes to ignore in the aiida hash
+            node_attributes: #mapping of entry points to list of attributes to ignore in hashing of nodes with those entry points
+                diff: [..]
+
+An example, where it might be necessary to modify the options in the ``ignore`` namespace, is testing workchains across multiple AiiDA core versions.
+When using a AiiDA archive created in version ``1.6`` and testing the workchain with this archive in version ``2.X`` calcjob nodes contain new metadata
```
ltalirz:
```suggestion
When using a AiiDA archive created in version ``1.6`` and testing the workchain with this archive in version ``2.X``, calcjob nodes contain new metadata
```

ltalirz (dismissed):

ltalirz:
Thanks @janssenhenning !

It appears I can't commit the suggested changes myself. Please do so. Afterwards I will merge

ltalirz (approved):

PR #68, opened 2023-01-31
fix: add option to disable MPI

ltalirz:
CalcJob plugins can specify default values for whether codes should use MPI or not: using the `metadata.options.withmpi` variable.

aiida-testing can run calculations with a single MPI process if an MPI library is installed. However, the MPI library is then also required in order to fetch the results from the test database (because the mock executable will be run with `mpirun -np 1 ...`).

Here we add an option `--mock-disable-mpi` that allows users to disable the use of MPI (e.g. in order to run tests on CI in conjunction with `--mock-fail-on-missing`).

### DIFF
--- a/aiida_testing/mock_code/_fixtures.py
+++ b/aiida_testing/mock_code/_fixtures.py
@@ -26,7 +26,6 @@ __all__ = (
     "mock_regenerate_test_data",
     "mock_fail_on_missing",
     "mock_disable_mpi",
-    "mock_disable_mpi_force",
     "testing_config",
     "mock_code_factory",
 )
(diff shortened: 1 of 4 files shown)


aiida_testing/mock_code/_fixtures.py
```diff
@@ -71,6 +79,44 @@ def mock_fail_on_missing(request):
...
+
+    from aiida.orm import CalcJobNode  # pylint: disable=import-outside-toplevel
+    monkeypatch.setattr(CalcJobNode, 'get_option', _mock_get_option)
+
+
+@pytest.fixture(scope='function')
+def mock_disable_mpi_force(monkeypatch):
+    """Enforce `withmpi=False` independent of `--mock-disable-mpi` cli option.
+
+    Mostly required to test the functionality.
```
greschd: If this is only for internal use, maybe we should rename it to `_mock_disable_mpi_force`?

greschd (dismissed):

ltalirz:
Cheers; thanks for the super-quick review, it actually doesn't work yet ;-)

Tests pass locally because I have an mpi library installed.
I'll need to monkeypatch slightly differently

greschd:
I'm guessing the Python 3.7 failure is because the latest AiiDA doesn't support that, so an older version is installed.
We should probably also drop support for 3.7 in a separate PR, and rebase on that.

ltalirz:
Cheers, I believe some of the archive cache fixtures cater to older AiiDA versions, so for the moment I will simply add some conditionals based on the version.

ltalirz:
Thanks for the quick input @greschd , much appreciated!

Issue #72, opened 2024-11-17
Release automation


danielhollas:
We should have an automated release workflow via GitHub actions, we can use the same thing as other repos in aiidateam.

After that I'd release 0.1 version for the current users to use. Any further (potentially breaking) changes then can go to 0.2 release.

Thoughts?

danielhollas:
Before publishing, we also want to change the name of the package, per discussion in #60. Looks like the discussion converged to `aiida-test-cache`.

ltalirz:
> Before publishing, we also want to change the name of the package, per discussion in #60. Looks like the discussion converged to `aiida-test-cache`.

Fine with me

danielhollas:
It looks like there are only 5 repositories with `aiida-testing` in their dependencies (either as a git dependency or via `aiidalab-testing-dev` package on PYPI). So we can easily create PRs for then once we publish under a new name. We should also somehow yank the `aiidalab-testing-dev` package if possible, and if not at least update its description to point to the new package.

danielhollas:
Should also configure automatic versioning on ReadTheDocs based on tags
https://docs.readthedocs.io/en/stable/guides/automation-rules.html

Issue #60, opened 2022-11-20
consider name change


ltalirz:
I am not familiar with the genesis of the `aiida-testing`, but it is obviously highly generic and does not convey what functionality the plugin adds on top of what already exists in `aiida-core`.

Some thoughts: The plugin focuses on 
 * tests of processes
 * tests that integrate AiiDA and the simulation software
 * tests that go from setting up calculation inputs through to the final result of the calculation/workflow

Name candidates could therefore be:
 1. `aiida-process-tests`
 2. `aiida-integration-tests`
 3. `aiida-e2e-tests` (e2e=end-to-end)

I would probably vote for option 3

greschd:
> I am not familiar with the genesis of the `aiida-testing`

AFAICR, it was born (mostly) out of the Feb'20 AiiDA Hackathon.

I agree, the name is a bit unfortunate. In particular since it could also be interpreted as "testing" channel for `aiida-core` features. I think it might also be because we might've intended to add more testing-related features, but never did.

Both features of the code relate to speeding up the tests via caching. So maybe, calling it something like `aiida-test-cache` would be appropriate?

It was at one point named `aiida-mock-codes`, but renamed in https://github.com/aiidateam/aiida-testing/commit/525ec6e7ba406f49b0cf3508aa67d32bafa78d4d when the export_cache was added.

ltalirz:
That's true, `aiida-test-cache` sounds good to me as well

greschd:
If we want to be even more specific, `aiida-test-cache-helpers` or `aiida-test-cache-utils` would also work for FMPOV.

@ltalirz feel free to implement the name change (whichever option you think fits best).

ltalirz:
@unkcpz (maybe also @chrisjsewell): you mentioned that you wanted to publish aiida-testing on PyPI.
This means, the time to decide on the name change is now.

I personally would vote for `aiida-test-cache` - I think it is specific and easy to understand.
If you agree, please rename the package before you publish.

unkcpz:
🤔I don't think aiida-test-cache is clear. I personally would vote for aiida-test-helper or more specifically aiida-unittest-helper.

ltalirz:
I would read `aiida-test-cache` as "a plugin that provides a cache for testing", which is exactly what the plugin does.
I guess you read it differently?

[...]

PR #73, opened 2024-11-17
Support new Python versions

danielhollas:
Test with new Python versions up to 3.12 (aiida-core doesn't support 3.13 yet).

EDIT: See comments below, in this PR I restricted the aiida-core versions to below or equal to 2.2, which is the last version that doesn't break any tests. Will open a follow-up PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,11 +53,6 @@ jobs:
     - name: Install package
       run: uv pip install --system ${{ matrix.editable_install_option }} .[testing]
 
-    - name: Install python dependencies
-      run: |
-        pip install --upgrade pip
-        pip install ${{ matrix.editable_install_option }} .[testing]
-
     - name: Run test suite
       env:
         # show timings of tests
@@ -74,7 +69,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: 3.11
+        python-version: 3.7
 
     - name: Set up uv
       uses: astral-sh/setup-uv@v3
@@ -95,7 +90,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: 3.11
+        python-version: 3.10
 
     - name: Set up uv
       uses: astral-sh/setup-uv@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
         version: ${{ env.UV_VER }}
 
     - name: Install package
-      run: uv pip install --system -e .[dev]
+      run: uv pip install --system -e .[pre_commit]
 
     - name: Run pre-commit
       uses: pre-commit/action@v3.0.1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
       fail-fast: false
 
     services:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: '3.11'
+        python-version: '3.12'
 
     - name: Set up uv
       uses: astral-sh/setup-uv@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
     - name: Set up uv
       uses: astral-sh/setup-uv@v3
       with:
-        version: ${UV_VER}
+        version: ${{ env.UV_VER }}
     - name: Install package
       run: uv pip install --system ${{ matrix.editable_install_option }} .[testing]
 
@@ -79,7 +79,7 @@ jobs:
     - name: Set up uv
       uses: astral-sh/setup-uv@v3
       with:
-        version: ${UV_VER}
+        version: ${{ env.UV_VER }}
     - name: Install package
       run: uv pip install --system -e .[docs]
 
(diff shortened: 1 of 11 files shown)


danielhollas:
The `archive_cache` tests are failing for AiiDA 2.6, which changed which objects are included in the hash:
- removed plugin and aiida-core versions from hash
- added node's `__class__` attribute in https://github.com/aiidateam/aiida-core/pull/6321

I'll investigate more tomorrow.

docs/source/conf.py
```diff
@@ -16,12 +14,7 @@
 import time
 import contextlib
 
-from aiida.manage.configuration import load_documentation_profile
-
 import aiida_testing
-
-load_documentation_profile()
```
danielhollas, line 23: This was removed since it is no longer needed in AiiDA 2.0

docs/source/conf.py
```diff
@@ -37,7 +30,6 @@
     with contextlib.suppress(ImportError):
         import sphinx_rtd_theme
         html_theme = 'sphinx_rtd_theme'
-        html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
```
danielhollas, line 40: Apparently no longer necessary, generated a warning

.github/workflows/ci.yml
```diff
@@ -9,12 +12,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python-version: ['3.9', '3.10']
-        editable_install_option: ['-e', '']
-        include:
-        #Test for aiida-core 1.X (since aiida-core 2.X does not support python 3.7)
-        - python-version: '3.7'
-          editable_install_option: '-e'
```
danielhollas, line 17: Let's just test non-editable install here, we install editably in pre-commit and docs jobs anyway.

danielhollas:
Unbelievable, all green for AiiDA 2.2. So version 2.3 broke the (ironically named) `test_broken_code`

```
FAILED mock_code/test_diff.py::test_broken_code - AssertionError: assert () == ('1,2c1', '< ...silly walks.')
  
  Right contains 5 more items, first extra item: '1,2c1'
  
  Full diff:
  + ()
  - (
  -     '1,2c1',
  -     '< Lorem ipsum dolor..',
  -     '<',
  -     '---',
  -     '> Please report to the ministry of silly walks.',
  - )
=================== 1 failed, 19 passed, 1 warning in 36.57s ===================
Error: Process completed with exit code 1.
0s
0s
```

I am putting this up for review, and will do a follow up fixes for 2.3 and 2.6 in separate PR so that the fixes are not drowned in unrelated changes here.

pyproject.toml
```diff
@@ -55,20 +59,13 @@ testing = [
 ]
 pre_commit = [
     "pre-commit",
-    "pylint~=2.12.2",
-    "mypy==0.930",
+    "pylint~=3.3.1",
```
danielhollas, line 62: I'm gonna get rid of pylint for ruff, but in a separate PR.

danielhollas: I begin to really appreciate the beauty of single rust binaries, this pylint pin was a reason why the installation for newer Python versions did not work. :-(

pyproject.toml
```diff
@@ -55,20 +59,13 @@ testing = [
...
 dev = [
-    "sphinx",
-    "sphinx-rtd-theme",
-    "pgtest~=1.3.1",
-    "aiida-diff",
-    "pytest-datadir",
-    "pre-commit",
-    "pylint~=2.12.2",
-    "mypy==0.930",
+    'aiida-testing[testing,pre_commit,docs]',
```
danielhollas, line 68: Self-referencing extras are allowed in `pyproject.toml` (but not in `setup.cfg`)

.github/workflows/ci.yml
```diff
@@ -9,12 +12,7 @@ jobs:
...
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
```
unkcpz, line 15: We can ask on the discourse and if no one using aiida-core < 2 with this anymore, maybe can drop the support for 3.7 and 3.8? 
I think if it is going to make a release, it is not a problem to drop the support and versioning it?

docs/source/conf.py
unkcpz: This was dropped at 2.5.0: https://github.com/aiidateam/aiida-core/blob/main/CHANGELOG.md#v250---2023-12-20

unkcpz (changes requested):
Thanks @danielhollas to look at this repo.
I remember now there was a discussion on the naming of the package at https://github.com/aiidateam/aiida-registry/pull/206

If we anyway want to change the name and make a release, I think we can start to support it from the latest version (or the nearest one) ?

.github/workflows/ci.yml
danielhollas: Let's have this discussion at #72. There I proposed to release the code basically as is (after renaming it) as version 0.1.0 so that existing users can easily switch to the official package. After that all breaking changes and droping support can be done in v0.2.

danielhollas: > one using aiida-core < 2 with this anymore, maybe can drop the support for 3.7 and 3.8?

btw: Dropping support for AiiDA v1 should anyway be done in a separate PR, since there will be a lot of code cleanup associated with that.

danielhollas:
@unkcpz thanks for taking a look!

I'd like this PR to be only about making the CI green and supporting new Python version. 

Regarding naming, the full discussion is in #60. Let's discuss release in #72.
Regarding dropping AiiDA v1 support, I'll do it in a separate PR since that's going to be larger cleanup.

unkcpz:
> I'd like this PR to be only about making the CI green and supporting new Python version.

I see, and I agree small PRs are betters. Would you mind open an issue and put those plans with checkbox? I guess you'll working on this during the coding week?

unkcpz (approved):

danielhollas:
Yep, the plan is to work on this during coding week. I'll open an issue with a plan.

danielhollas:
Before I merge this I'll try to fix the tests for newer AiiDA versions. If it is not too hard I'll do it in this PR.

danielhollas:
I'll merge this, opened #75 as follow up for tomorrow.

PR #76, opened 2024-11-18
Rename package: `aiida-testing` -> `aiida-test-cache`

### LINKED ISSUE #60: consider name change

danielhollas:
Fixes #60.

There are two things to discuss / bikeshed here:

1. `aiida-test-cache` or `aiida-testcache`?
2. Name of the default config file. I think we should definitely rename it to be consistent with the new name of the package, but do we also want to support the old name in a backwards compatible manner? My vote is no to keep the code simple, given that the number of current users is very low and I can simply make a PR to all of them.

Before merging this (or after?) I also need to:
- [x] Rename the repo
- [x] Change the configuration on ReadTheDocs appropriatelly

I also fixed the documentation build, see the new build for this PR https://aiida-testing--76.org.readthedocs.build/en/76/index.html


pyproject.toml
```diff
@@ -51,7 +51,7 @@ dependencies = [
 
 [project.optional-dependencies]
 docs = ["sphinx", "sphinx-rtd-theme"]
-testing = [
```
danielhollas, line 54: While I was here, I made it in line with other AiiDA repos that use `tests` extras.

unkcpz (approved):
Looks all good to me.

danielhollas:
@unkcpz thank you very much for review! 

Renaming done, let's go! :rocket:

PR #81, opened 2024-11-20
Rename default config file to `.aiida-test-cache-config.yml`

danielhollas:
...to align the name to the new name of the package.

We could also support the old name for backwards compatibility but I don't think it's worth it the additional code complexity. The existing users will anyway need to change the name of the package in their dependencies, so renaming one file is not much extra work.

@PhilippRue it would be super helpful if you could test this in case you have capacity. Thank you!


PhilippRue:
> We could also support the old name for backwards compatibility but I don't think it's worth it the additional code complexity. The existing users will anyway need to change the name of the package in their dependencies, so renaming one file is not much extra work.

I would also say we don't need to support the old name.
 
> @PhilippRue it would be super helpful if you could test this in case you have capacity. Thank you!

Works 👍

PR #84, opened 2024-11-29
Support aiida 2.5

### LINKED ISSUE #75: aiida-core v2.3 broke test_broken_code

danielhollas:
We saw a failing mock_code test since aiida 2.3.0 (see #75)
This was caused by just a whitespace change in the AiiDA
submission script (controlled by the direct Scheduler plugin).

Concretely, there was an extra space at the end of the last
line of the script (the executable line) before 2.3.
As a short term solution to preserve the existing hashes,
we append a space if the script doesn't end with one.

This change is purely for backwards compatibility and should be
reverted right after 0.0.1 release.

As a longer term solution we should make this less brittle, perhaps by removing all white space in the submission script. But even that still makes us dependent on the core.direct scheduler implementation. I'll open a separate issue to track that.

Closes #75


ltalirz (approved):
well done figuring this one out

PR #88, opened 2024-12-04
Add code coverage

danielhollas:


danielhollas:
Hi @agoscinski,

I'd like to setup Codecov for this repo, but right now the uploads are failing since I don't have the Codecov token. Rather then generating the token specific to this repo, we could instead either.

1. Not require tokens for uploads (in the past Codecov required them, but now they do not. 
2. Setup a global token for the `aiidateam` organization and add it as a organization-level secret (this is what we have setup for AiiDAlab I believe. The settings should be on [this page](https://app.codecov.io/account/gh/aiidateam/org-upload-token)

image

It looks like the tokenless approach is the easiest here (requires codecov/action v5, more info here:
https://docs.codecov.com/docs/codecov-tokens#uploading-without-a-token

Thanks for your help

agoscinski:
I added for now an organization secret `CODECOV_TOKEN` from code cov. I don't want to deactivate the requirement for tokens before I have talked with some other people about it.

PR #89, opened 2024-12-05
Remove deprecated `ignore_files` arg in `mock_code_factory` fixture

danielhollas:


ltalirz (approved):
Yeah, I don't remember why we had both `ignore_files` and `ignore_paths`. I guess `ignore_paths` is sufficient

`ignore_files` is used in some places though, so they will need to adapt. eg

https://github.com/lsmo-epfl/nanoporous_screening
https://github.com/greschd/aiida-tbextraction
https://github.com/anoopkcn/aiida-fleur

danielhollas:
@ltalirz thanks!

> ignore_files is used in some places though, so they will need to adapt. eg

Yeah, I am planning to submit PRs for current users to migrate to aiida-test-cache package, so I can change those as part of that.

